PR #64, opened 2019-02-06
ENH: FormHandler edits on files should infer type

pratapvardhan:
## Summary
FormHandler edits on files/dataframes should infer type from original data.

```python
import gramex.data

url = 'https://learn.gramener.com/guide/formhandler/flags?_format=csv'
df = pd.read_csv(url)
# ?Continent=Asia&c1=101   --  handler.args will also store values as strings.
count = gramex.data.update(df, args={'Continent': ['Asia'], 'c1': ['101']}, id=['Continent'])
```

*Currently,* this would store `c1` with string `'101'`

This PR fixes it.


sanand0:
Closed via 66967472

PR #70, opened 2019-02-11
ENH: --watch=False / init(watch=False) disables autoreloading gramex.yaml

pratapvardhan:
- Allows user to disable watching `gramex.yaml`, which disables autoreloading `gramex.conf` via `gramex --nowatch`
- Also, provides `gramex.init(watch=True)` to explicitly not watch files.

### DIFF
--- a/gramex/__init__.py
+++ b/gramex/__init__.py
@@ -264,15 +264,16 @@ def init(force_reload=False, **kwargs):
     from . import services
     globals()['service'] = services.info    # gramex.service = gramex.services.info
 
-    # Set up a watch on config files (including imported files)
-    from services import watcher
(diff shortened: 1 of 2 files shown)


sanand0:
@pratapvardhan Let's discuss when we speak next, please. I'd like to understand the implications.

sanand0:
- Add `app.watch: True` to gramex/gramex.yaml (instead of `nowatch`)
- Use `gramex.init({'app': {'watch': False}})` instead of `gramex.init(watch=False)`.
  This will be available directly in `final_config`
- Later on, refactor `gramex.services.url()` to use a function that updates individual URL specs

PR #71, opened 2019-02-11
BUG: FormHandler edits should adhere to existing types

pratapvardhan:
Ref to bug introduced due to !64


sanand0:
@pratapvardhan -- agreed on the change. Let's merge when we speak this week.

PR #127, opened 2019-03-15
ENH: Lazy load spaCy models

jaidevd:
Spacy models were mostly globals in NLG modules. This PR adds them as function parameters wherever required and loads a model only once.

### DIFF
--- a/gramex/apps/nlg/grammar.py
+++ b/gramex/apps/nlg/grammar.py
@@ -1,14 +1,14 @@
 from inflect import engine
 from tornado.template import Template
 
-from gramex.apps.nlg import utils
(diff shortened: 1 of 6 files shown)


gramex/apps/nlg/grammar.py
```diff
@@ -3,15 +3,15 @@
 from spacy.lemmatizer import Lemmatizer
```
sanand0: Please lazy-load spacy. To test it, uninstall spacy and run the guide. It should still work.

jaidevd:
@sanand0 Summary of my commits post your review:
1. No spacy attribute is imported by default
2. Everything is conditionally imported if it's not present in a local dict.
3. Removed gramex's dependency on spacy - moved it to the NLG app.

sanand0:
Merged via c161a6a0

PR #133, opened 2019-03-21
BUG: markdown meta should be cached

pratapvardhan:
Markdown frontmatter should be cache along.


sanand0:
Looks good. Will merge

Issue #96, opened 2019-02-22
gramex.data.insert should create primary keys


CalmDownKarm:
Gramex.data.insert currently creates tables/files if they don't exist. 
if it's creating a table - and there's an id column specified, it should apply a primary key constraint on that column in the table it creates.

CalmDownKarm:
@s.anand @pratapvardhan since the new table is created via a pandas.to_sql call, 2 potential ways to fix - 
either overwrite pandas.to_sql as in this stack overflow answer https://stackoverflow.com/questions/30867390/python-pandas-to-sql-how-to-create-a-table-with-a-primary-key 
or manually check if the table exists and write an SQL query to create it/ alter the table and add a primary key (which would fail for sqlite and possibly other dbs). 
We could also potentially create a new table with a primary key, copy data over and drop the old table. 
None of these solutions are very elegant. Do you have any suggestions?

sanand0:
Suggested approach: add this before the `.to_sql()`

```
if table does not exist:
    con.execute(pd.io.sql.get_schema(data, keys=our-primary-key))
```

CalmDownKarm:
Feature is currently broken for python 2 if table names are unicode strings.

PR #136, opened 2019-03-22
adding support for primary key creation for data.insert

CalmDownKarm:
issue #96


sanand0:
@CalmDownKarm -- `pd.io.sql.get_schema(rows, name=table, keys=id, con=engine)` is brilliant!

I've refactored the test cases a bit. Will merge once this passes.

PR #137, opened 2019-03-23
BUG: rename NLG modules to avoid conflict

sanand0:
@jaidevd Gramex adds app paths to sys.path. This means that module names like `utils` are importable via "import utils". But many apps already do this. For example, gramex/tests uses a utils.py. uat.gramener.com has a few apps that use utils.py. 

I've sent a PR that renames utils.py to nlgutils.py, and a few other modules. Is this fine? If yes, and if it doesn't break other stuff, please merge into dev.

### DIFF
--- a/gramex/apps/nlg/__init__.py
+++ b/gramex/apps/nlg/__init__.py
@@ -1,1 +1,1 @@
-from .search import templatize  # NOQA: F401
+from .nlgsearch import templatize       # NOQA: F401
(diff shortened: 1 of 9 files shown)


jaidevd:
LGTM, @sanand0 - but I'm not authorized to merge this.

sanand0:
Oh, OK. I'll merge. Thanks @jaidevd

Issue #145, opened 2019-04-08
FormHandler support aggregations without groupby


pratapvardhan:
For example you might like to get `max` , `min`, `avg` of column `c1` using `flags?_c=c1|min&_c=c1|avg&_c=c1|max`  perhaps?  with `?<col><|><agg>` `|` being the separator.

```
results = df.agg({'c1': [min, max, 'mean']}).T

query = "SELECT MIN(c1), MAX(c1), AVG(c1) FROM "table_name"
```

```
  min   max       mean
 0.0  97.0  29.795918
```

Would be useful for quick explorations with formhandler.  Thoughts?

sanand0:
Spec-wise, we can treat `?_by=` as a group-by on a constant, and it aggregates all other columns.

Could you share some real-life examples where you would use this, @pratapvardhan?

pratapvardhan:
Quite didn't get the first part. Also want to get values from other columns, "SELECT MIN(c1), MAX(c2), AVG(c3) FROM "table_name"``

I'd use them for KPIs, summarized values for narratives. Let's discuss when we speak next.

sanand0:
I meant that `flags?_by=_c=c1|min&_c=c1|avg&_c=c1|max` will do what you want.

PR #150, opened 2019-04-11
ENH: FormHandler support column aggregations. See #145

pratapvardhan:
See #145

### DIFF
--- a/gramex/data.py
+++ b/gramex/data.py
@@ -706,38 +706,40 @@ def _filter_frame(data, meta, controls, args, source='select', id=[]):
         # Apply controls
         if '_by' in controls:
             by = _filter_groupby_columns(controls['_by'], data.columns, meta)
+            # If ?_c is not specified, use 'col|sum' for all numeric columns
+            # TODO: This does not support ?_c=-<col> to hide a column
+            col_list = controls.get('_c', None)
+            if col_list is None:
+                col_list = [col + _agg_sep + 'sum' for col in data.columns      # noqa
+                            if pd.api.types.is_numeric_dtype(data[col])]
+            agg_cols = []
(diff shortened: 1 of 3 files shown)


gramex/data.py
```diff
@@ -829,29 +831,28 @@ def _filter_db(engine, table, meta, controls, args, source='select', id=[]):
         # Apply controls
         if '_by' in controls:
             by = _filter_groupby_columns(controls['_by'], colslist, meta)
-            if len(by) > 0:
-                query = query.group_by(*by)
```
pratapvardhan: For SQL, just removed `len(by) > 0` check.

gramex/data.py
```diff
@@ -706,38 +706,40 @@ def _filter_frame(data, meta, controls, args, source='select', id=[]):
...
-                    col, agg, val = _filter_col(key, data.columns)
-                    if agg is not None:
-                        # Convert aggregation into a Pandas GroupBy agg function
-                        agg = agg.lower()
-                        agg = _frame_functions.get(agg, agg)
-                        agg_cols.append(key)
-                        if col in agg_dict:
-                            agg_dict[col].append(agg)
-                        else:
-                            agg_dict[col] = [agg]
```
pratapvardhan: Moved this chuck before `len(by) > 0` and `data.groupby(by)` to later.

pratapvardhan:
Looks to  be unrelated failing scheduler test case.

sanand0:
- Add test case for `{'_by': []}` without aggregations
- Add test case for `{'_by': [], '_c': []}` and `{'_by': [], '_c': ['']}`. Check that it does not throw an exception. Do not test return value

pratapvardhan:
@sanand0 -- tests are added.

Issue #81, opened 2019-02-15
docker packaging


CalmDownKarm:
Switch the docker image to use miniconda instead of anaconda - this should reduce the size of the compiled images considerably.

Also perhaps we should deploy enterprise and dev tags to dockerhub as well?

Warlord77:
Can I pick up this issue ?

sanand0:
Sure @Warlord77. Perhaps we could speak over a call? My number is +91 9741 552 552, and I'm reachable on mobile until tomorrow, and on WhatsApp after that.

PR #151, opened 2019-04-16
Dockerfile updated to miniconda

CalmDownKarm:
Built image is 2.7Gb, roughly half the size of the Anaconda based image. 

partially solves #81 @sanand0 should I add images for gramexenterprise and gramex dev?

### DIFF
--- a/pkg/docker-py3/Dockerfile
+++ b/pkg/docker-py3/Dockerfile
@@ -20,8 +20,7 @@ RUN apt-get install -y python-dev python3-dev build-essential
 # ModelHandler Requires scikit-learn
 RUN conda install scikit-learn
 # Install gramex
-RUN pip install psycopg2-binary && \
-    pip install gramex && \
+RUN pip install gramex && \
     gramex setup --all && \
     gramex license accept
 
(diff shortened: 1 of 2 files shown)


sanand0:
Let’s plan only for the latest released versions Karmanya. The dev branch
changes too often. For enterprise edition, we don’t want a public Docker
installable.

I will review this shortly.

CalmDownKarm:
@sanand0 based on our phone conversation
- Added instructions to the readme inside this directory about the tests and which ones fail. Despite having mysql and postgres in the container, data.py db related tests all get skipped - need to figure out why. 
- In terms of the packages that should be installed via conda, lineprofiler and rpy2 I think we can skip for the moment, they aren't very popular features of gramex, and someone that needs them can always create a custom Dockerfile. Sklearn I would include in this list as well, but since the guide will currently break without it, I've included in the list.

sanand0:
@CalmDownKarm I'll merge this once you create a test script as mentioned in #152. It can be a manually run script for now. You can add it to this PR if you like. I'll validate it, run it, and merge. Thanks!

PR #167, opened 2019-06-03
ENH: Chrome capture supports device emulation. Upgrade puppeteer ^1.17.0

pratapvardhan:


pratapvardhan:
Note: Test case is failing due to version upgrade perhaps. Works fine on my machine though.

sanand0:
get_layout_elements returns different values on Windows and Linux. Hence "Header" may be in `layout[0]` (Linux) or `layout[1]` (Windows). 

Your discretion on whether to provide a quick fix (e.g. check either in `[0]` or `[1]`) or really solve the problem. But either way, please add a comment to `test_capturehandler.py` about this)

pratapvardhan:
Turns out, pdf is rendered differently by puppeteer on Windows and Linux.

Issue #170, opened 2019-06-05
Deprecate datahandler, queryhandler and remove it's dependencies


pratapvardhan:
datahandler, queryhandler are not longer used. Blaze, odo are not maintained too. Let's remove them and add deprecation notes to guide (remove config, just add note of removal).

pratapvardhan:
Fixed via 8b37e803acf18ef2b41b95e103f520f11fd370c7

PR #172, opened 2019-06-06
DEP: Remove DataHandler, QueryHandler #170

pratapvardhan:
See #170


sanand0:
Good to go on all counts except `formutils.py`. Can we somehow eliminate it and replace it with a single line of Python in the schedule function itself? Sort of like a `pd.read_csv(url).to_sql(db)`? If nothing is obvious after 15 minutes of thought, please merge.

pratapvardhan:
Yes, removed `formutils.py`. Using

```
schedule:
  data-setup-flags-db:
    function: pandas.read_csv(r'$YAMLPATH/flags.csv').to_sql(
      'flags',
      sqlalchemy.create_engine('sqlite:///$YAMLURL/database.sqlite3'),
      if_exists='replace', index=False)
```

PR #173, opened 2019-06-06
ENH: Replace chart styled dynamic columns name and table styled dynam…

nivedita-deshmukh:
Replace chart styled dynamic columns name and table styled dynamic columns name with URL args value

Purpose:
Data columns name will be dynamic each time on which need to pass CSS style.
so for that in .yaml fill will be passing a static name and which can be replaced with dynamic columns name via URL args
so the actual data columns name can be displayed as table column name and chart axis label in the output PPT.


pratapvardhan:
@nivedita-deshmukh -- I believe this is resolved with column positions? If so, could you close this PR? If not, could you add the usage spec in description?

Issue #171, opened 2019-06-05
Migrate to sklearn.externals.joblib to joblib in gramex.ml


pratapvardhan:
Warning message states

`sklearn.externals.joblib` is deprecated in 0.21 and will be removed in 0.23. Please import this functionality directly from joblib, which can be installed with: pip install joblib. If this warning is raised when loading pickled models, you may need to re-serialize those models with scikit-learn 0.21+.

- We need to add `joblib` to miniconda setup
- Test old pickles work, else recreate

pratapvardhan:
Closed via #175

PR #175, opened 2019-06-07
sklearn 0.21 onwards use joblib to persist models #171

pratapvardhan:
…


sanand0:
All changes are fine, @pratapvardhan -- except `import os.path` is not required if we `import os`.

Please merge

PR #177, opened 2019-07-01
Doc: Standardize quickstart and dashboard tutorials

### LINKED ISSUE #174: Feedback on tutorials/charts
Comments on 9ef6ba74

- Add an "Outcome" header to each tutorial, indicating the outcome
- Step 1: show the tabular output of the modify: result so people get a feel for the out (?)
  - Maybe by directly embedding a part of the output in-place
  - like templates/imdb-template/#load-the-data maybe
- Step 3: Merge the two JavaScript snippets into one block
- Step 3:
  - g1.url.parse(baseDataURL).toString() == baseDataURL. Replace it
  - Don't use e.hash.relative. You probably want e.hash.{hash|search|searchKey}
  - I suspect `spec.data.url = baseDataURL + '#' + e.hash.hash` might work
- Step 4:
  - Move the `draw_chart()` part above, explain that it's vega view event (and maybe a line about the) first
  - Then explain what we do in the event handler
  - filterTableOnClick is too complex. Simplify it. And then add comments to explain

jaidevd:
Closes #174


pratapvardhan:
Feedback:

- Since content is progressive for tutorials. Use server templates and reuse them in `chart1.html` `chart2.html` and likewise for `index1.hml` etc.

- Or use `chart(.*).html` pattern and use `handler.path_args[0]` to conditionally load content?

jaidevd:
@pratapvardhan I don't follow. What do you mean by server templates?

jaidevd:
Closing and moving to https://github.com/gramexrecipes/gramex-guide/pull/2

Issue #179, opened 2019-07-18
gramex.data.filter does not cache sql


pratapvardhan:
Internally, in `data.py` we are using
```
data = gramex.cache.query(text(query), engine, state, params=all_params)

# elsewhere in cache.py (text(query), engine) is used for cache lookup

def query(sql, engine, state=None, **kwargs):
    key = (sql, engine.url)
    current_status = _cache.get(key, {}).get('status', None)
```

Now, `text(query)` returns a new object for same `query` text. Hence, internal caching of this key doesn't work.

Probably use a cache lookup for `text(query)` or internally in `cache.py` use sql text for key instead?

sanand0:
@pratapvardhan Let's discuss when we speak next, please?

pratapvardhan:
fixed via df97890

PR #180, opened 2019-07-29
BUG: Serialize query cache key #179

pratapvardhan:


sanand0:
Looks good @pratapvardhan -- please take a look at the Travis fails and merge

PR #182, opened 2019-07-29
ENH: Change shape's style or properties based on data val.

nivedita-deshmukh:
Enhancement features:
1. PPT Shape's style or properties can be modified based on given data values from a python function (dynamically property change support).
2. Added new feature Shape rotation to rotate a shape with the given data value.

From the ".Yaml" file input definition will be -
```
change:
  Shape1:
    data:
      function: risk_deck_pptx.test_py_func(data['shape_test'], handler)
    css:
      style:
        fill: '{{ fill }}'
        stroke: '{{ stroke }}'
        rotation: '{{ rotation }}'
```

### DIFF
--- a/gramex/pptgen/commands.py
+++ b/gramex/pptgen/commands.py
@@ -157,12 +157,15 @@ def generate_style(style, data, handler):
 
 def rect_css(shape, **kwargs):
     '''Function to add text to shape.'''
-    for key in {'fill', 'stroke'}:
+    for key in {'fill', 'stroke', 'rotation'}:
         if kwargs.get(key):
-            fill = shape.fill if key == 'fill' else shape.line.fill
-            rectcss = kwargs[key].rsplit('#')[-1].lower()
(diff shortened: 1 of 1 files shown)


pratapvardhan:
@nivedita-deshmukh -- could you add test case to understand the usecase better? Or share the reproducible example? input pptx, output pptx, same data -- attach these in the comment, if you're unable to add a test case.

nivedita-deshmukh:
PFA of test data and ppt on the below link

[https://drive.google.com/open?id=1WC9YbCv1R1oL0WR090qBbX8c8VlLZ5tU](url)

nivedita-deshmukh:
@pratapvardhan please do let me know if anything else more require.

pratapvardhan:
@nivedita-deshmukh would explore [Register a new command to PPTGen](https://learn.gramener.com/guide/pptxhandler/#register-register-a-new-command-to-pptgen) as an alternative.

PR #183, opened 2019-07-31
ENH: Chart data columns positional identifiers changes.

nivedita-deshmukh:
**Purpose: Using column position as well assign the color to a column from ".yaml" definition.**

Here is the sample ".yaml" example

```
chart_name:
  chart:
    data:
      function: >
        py_file.py_func(data['test_data'], handler)
          x: 'label'
            color:
              2: '#B9CDE5'
              3: '#558ED5'
```

### DIFF
--- a/gramex/pptgen/commands.py
+++ b/gramex/pptgen/commands.py
@@ -339,6 +339,17 @@ def chart(shape, spec, data):
             style[key] = compile_function(style, key, data, handler)
 
     data = compile_function(info, 'data', data, handler)
+
+    """chart data columns positional identifiers changes."""
+    dt_cols = data.columns
+    input_colspec = spec['chart'].get('color', {})
+    input_cols = input_colspec.keys()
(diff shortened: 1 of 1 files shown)


pratapvardhan:
@nivedita-deshmukh -- could you add test case to understand the usecase better? Or share the reproducible example? input pptx, output pptx, same data -- attach these in the comment, if you're unable to add a test case.

nivedita-deshmukh:
@pratapvardhan PFA of test data and ppt on the below link
[https://drive.google.com/open?id=1WC9YbCv1R1oL0WR090qBbX8c8VlLZ5tU](url)

nivedita-deshmukh:
@pratapvardhan please do let me know if anything else more require.

pratapvardhan:
@nivedita-deshmukh would explore [Register a new command to PPTGen](https://learn.gramener.com/guide/pptxhandler/#register-register-a-new-command-to-pptgen) as an alternative.

PR #186, opened 2019-08-06
REF: Move guide, install on setup

pratapvardhan:
Files to see

- gramex/apps.yaml 
- setup.py
- testlib/test_setup.py
- add dependencies for latest miniconda / pandas


sanand0:
All fine. But in test_setup.py, please don't change exists(), and don't check for the guide. Those tests can be deleted.

Please create a new PR that removes the guide installation. If people run Gramex without gramex.yaml, treat it as an empty gramex.yaml (or anything else that makes the code elegant = smaller)

PR #190, opened 2019-08-10
REF: add joblib dependency, load sklearn on demand #188

### LINKED ISSUE #188: Add joblib as required dependency
Also, consider moving `from sklearn.preprocessing import StandardScaler` in `gramex/ml.py` inside `tran()` function.

pratapvardhan:
Fix #188

### DIFF
--- a/gramex/ml.py
+++ b/gramex/ml.py
@@ -1,14 +1,13 @@
 import os
 import six
+import json
 import inspect
 import threading
-import pandas as pd
-import json
 import joblib
+import pandas as pd
 from tornado.gen import coroutine, Return, sleep
 from tornado.httpclient import AsyncHTTPClient
 from gramex.config import locate, app_log, merge, variables
-from sklearn.preprocessing import StandardScaler
 
 # Expose joblob.load via gramex.ml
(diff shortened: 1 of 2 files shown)


jaidevd:
@pratapvardhan @sanand0 

1. There are a bunch of dependencies marked `# OPT`, which IMO should be removed, especially those related to DBs like pymysql or psycopg2. Having them means that I can't use gramex without having mysql or postgres installed.
2. I suggest we stop assuming that miniconda is installed. Often we install gramex in blank venvs which have none of the libraries (like requests, for e.g.) which gramex doesn't mention in `setup.py`.

sanand0:
PR is good to go.

@pratapvardhan -- please check if the DBs are necessary to install these on Windows + Linux. If yes, then we should remove these as dependencies. If not, let's discuss it. Could you speak to @jaidevd and raise this as a separate issue after this research.

@jaidevd re: Miniconda, it appears to be necessary. I don't think `pip install sklearn` works on Windows, for example. If you can validate that all required libraries work without conda, we can consider this. Will leave it to you to test and confirm, please.

PR #191, opened 2019-08-14
UPD: reading yaml with defaults

pratapvardhan:


gramex/cache.py
```diff
@@ -116,6 +116,13 @@ def _markdown(handle, **kwargs):
     return markdown(handle.read(), **{k: kwargs.pop(k, v) for k, v in _markdown_defaults.items()})
 
 
+@opener
+def _yaml(handle, **kwargs):
+    import yaml
+    defaults = {'Loader': yaml.FullLoader}
+    return yaml.load(handle.read(), **{k: kwargs.pop(k, v) for k, v in defaults.items()})
```
sanand0: @pratapvardhan Isn't this the same as `yaml.load(handle.read(), **defaults)`? Couldn't we just do:

```python
def _yaml(handle, **kwargs):
    return yaml.load(handle.read(), Loader=yaml.FullLoader)
```

pratapvardhan: @sanand0 we couldn't pass custom `Loader` like `gramex.cache.open(path, 'yaml', Loader=AttrDictYAMLLoader)` so we need to use `defaults` values and override with `kwargs`

PR #194, opened 2019-09-04
More text size utilities

bkamapantula:
## Current text size utils

`sm1`, `sm2`, `sm3`, `sm4`, `sm5`, `sm6`

## Proposed utils

`sm1`, `sm2`, `sm3`, `sm4`, `sm5`, `sm6`, `sm7`, `sm8`, `sm9`

This is inline after the discussions with design and UI teams.


gramex/apps/ui/bootstrap-theme.scss
```diff
@@ -264,12 +264,15 @@ $line-height-4: 2.0 !default;
...
+  sm1: 90%,
+  sm2: 80%,
+  sm3: 70%,
+  sm4: 60%,
+  sm5: 50%,
+  sm6: 40%,
+  sm7: 30%;
+  sm8: 20%;
+  sm9: 10%;
 );
```
pratapvardhan: This would be backward incompatible. We might want to think of having it support existing sm1-6

bkamapantula: agreed, retained the old classes in 2d9af68.

renamed the new utils with two overlapping classes for `80%` (`sm1` and `fsm2`) and `50%` (`sm3` and `fsm5`). retaining the redundant classes for naming convention uniformity and consistency when we deprecate the old utils.

pratapvardhan: Wondering if it should be `fs1: 10%, fs2: 20%, ...` proportional to number, instead of going reverse like sm. @sanand0 your preference?

bkamapantula: yes, [opacity](https://learn.gramener.com/guide/uicomponents/#text-opacity) utils are named similarly. updated to `fs*`

PR #200, opened 2019-09-21
pre-commit configuration in gramex init

bkamapantula:
This PR Introduces configuration for build error checks locally using `pre-commit` python package.

-  `pre-commit` is triggered on every `git commit` by default
- these checks are configured in `.pre-commit-config.yaml`: trailing whitespace, ending a file with single empty line, ensures larger than 500KB files aren't committed, valid YAML config, python checks via `flake8`, `JSON` checks 

# Test
- create a new directory
- run `gramex init` on terminal (*nix-based systems) or on `git bash` (Windows)

## Test Output

image

# Test in an existing live application
- Change a file (introduce a bug: for example, add a `console` statement in JS file) and run: `git commit`. Commit should fail. Fix the bug and re-commit. Commit should succeed.

### DIFF
--- a/gramex/apps/init/.pre-commit-config.yaml
+++ b/gramex/apps/init/.pre-commit-config.yaml
@@ -1,0 +1,20 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.0.0
+  hooks:
+  - id: trailing-whitespace
+    description: This hook trims trailing whitespace.
+  - id: end-of-file-fixer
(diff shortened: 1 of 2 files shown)


bkamapantula:
Sure @pratapvardhan, let's move it to a repo

pratapvardhan:
@bkamapantula I believe next steps could be

- Move init app to https://github.com/gramexrecipes/gramex-init or https://github.com/gramener/gramex-init ?
- Add an entry to gramex/apps.yaml like https://github.com/gramener/gramex/blob/dev/gramex/apps.yaml#L3
- Add a test case, and test in clean env / docker?
- Check if gramex-init upstream updates are synced back?

bkamapantula:
thanks @pratapvardhan, will try this and update in a week.

PR #217, opened 2020-02-13
ENH: FormHandler groupby URL allows retrieve specific groupby columns

tejesh0:


tejesh95:
@pratapvardhan If you are okay with `_by_c` spec, I will add tests and docs please.

tejesh95:
http://localhost:9988/api?_by=city&_by=product&_c=city&_c=sales|SUM&_format=html
would give a response with both `product` and `city` columns, overriding `_c=city`
image



http://localhost:9988/api?_by=city&_by_c=product&_c=sales|SUM&_format=html

would give a response with only `product` column ignoring `city`
image

Allowing FormHandler spec to follow a strict SQL group by semantics

PR #228, opened 2020-05-07
DriveHandler & FileManager

### LINKED ISSUE #222: Create DriveHandler as a FormHandler compatible UploadHandler
- Create UploadHandler option `formhandler: true`
- If `formhandler: true` store in a table called `files` (not `store`) with:
  - file
  - size
  - mime
  - filename
  - key
  - created
  - user.id -> user_id
  - user.<etc> -> user_<etc>
  - data.<etc> -> <etc>
- GET returns the data as a FormHandler rather than an object. Supports pagination, etc

Notes:
- The `store` table will be used to hold data in the old schema. The `files` table uses the new schema
- The schema will be altered when the handler is reloaded. If the schema is different, the new schema will be applied. No old fields will be deleted.
- Remove the migration from HDF5 store `_migrate_h5`

jaidevd:
Closes #222 

FileManager features:
- [x] importable like the admin app
- [x] listing, creating and deleting files
- [x] downloading files
- [x] drop files anywhere in the FormHandler table

ToDo:
- [x] Editing filenames and custom fields
- [x] Creating a `$(el).filemanager(opts)` function, extending `g1.formhandler`

What happens on failure? Of any kind? Can we simulate the following kinds of failures (at least) and test:
- [x] User uploads too large a file
- [x] User uploads file without permission
- [x] User uploads wrong type of file
- [x] Disk is out of space
- [x] Gramex stopped - wrong port

    
The right error message needs to be displayed.


gramex/apps/filemanager/gramex.yaml
```diff
@@ -0,0 +1,28 @@
...
+  apps/filemanager/home-$*:
+    pattern: /$YAMLURL/
+    handler: FileHandler
+    kwargs:
+      path: $YAMLPATH/index.html
+      template: true
+      filemanager_kwargs: $FILEMANAGER_KWARGS
+      auth: $FILEMANAGER_AUTH
+
+  apps/filemanager/help-$*:
```
sanand0, line 16: Keep help separately in help.html. Keep home page separately in index.html. Share the scaffolding through snippets or sub-templates.

PR #265, opened 2020-07-21
ElasticSearch logging

sanand0:


gramex/config.py
```diff
@@ -840,3 +842,55 @@ def used_kwargs(method, kwargs, ignore_keywords=False):
...
+
+def log(**kwargs):
+    '''
+    Writes the log into Elastic Search and application log.
+    Usage:
+        log(level='INFO',x=1, y=2, msg='log string') writes to ES as below
+        {'level': 'INFO', 'x': 1, 'y': 2, 'msg': 'log string', 'time': '2020-07-21 13:41:00', 
+            'port': 9988}
+    '''
+    log_args = kwargs.copy()
```
sanand0, line 883: No need to copy -- we're not changing any existing objects. kwargs are created dynamically

gramex/config.py
```diff
@@ -840,3 +842,55 @@ def used_kwargs(method, kwargs, ignore_keywords=False):
...
+    if 'level' in log_args.keys():
+        log_level = log_args['level'].upper()
+        if not log_level in allowed_log_levels:
+            log_args["level"] = "UNKNOWN"
+        log_args["level"] = log_level
```
sanand0, line 888: The above logic can be simplified. 

```python
log_level = log_args.get('level', 'UNKNOWN').upper()
log_args['level'] = allowed_log_levels.get(log_level, 'UNKNOWN')
```

Also, please use single quotes, not double-quotes

gramex/config.py
```diff
@@ -840,3 +842,55 @@ def used_kwargs(method, kwargs, ignore_keywords=False):
...
+    log_args["time"] = strftime("%Y-%m-%d %H:%M:%S", localtime())
+    log_args["port"] = app_log_extra['port']
+    app_log.info(log_args)
+    try:
+        es = ElDb().get_connection()
+        es.index(index=esl_conf['index'], body=log_args)
+    except Exception:
+        app_log.error("unable to store in ES")
```
sanand0, line 896: It would be good to call ElasticSearch only if it's defined in the config. So, we can have two functions for `log` -- e.g. `log` and `eslog`. If the ES logging is enabled, `eslog` will replace `log`.

Secondly, we don't want to push to ES synchronously. That can slow things down dramatically. Best to run this as a separate thread. We may also want to queue these and push them in a batch if required.

gramex/config.py
```diff
@@ -840,3 +844,78 @@ def used_kwargs(method, kwargs, ignore_keywords=False):
...
+            es = ElDb().get_connection()
+            es.index(index=variables['ES_INDEX'], body=log_doc)
+        except Exception:
+            app_log.error('unable to store in ES')
+
+if esl_conf['enabled']:
+    '''
+    Initialize a thread only if eslog is enabled in gramex.yaml
+    '''
+    est = threading.Thread(target=log_to_es, daemon=True)
```
sanand0, line 901: Use `gramex.threadpool` instead

gramex/gramex.yaml
```diff
@@ -216,6 +216,8 @@ log:
 eventlog:
     path: $GRAMEXDATA/logs/events.db          # Location of application event logs.
 
+eslog:
+    enabled: True
```
sanand0, line 220: This will automatically call `eslog(enabled=True)` in `gramex/services/__init__.py`. You can perform all initialization there.

gramex/config.py
```diff
@@ -840,3 +844,78 @@ def used_kwargs(method, kwargs, ignore_keywords=False):
             target = used if key in set(argspec.args) else rest
             target[key] = val
     return used, rest
+
+
+esl_paths = AttrDict()
+esl_paths['source'] = Path(__file__).absolute().parent
+esl_conf = (+PathConfig(esl_paths['source'] / 'gramex.yaml')).get('eslog', AttrDict())
```
sanand0, line 851: Move this into `gramex/services/__init__.py` under `def eslog(conf)`. You'll get the config directly

gramex/config.py
```diff
@@ -840,3 +844,78 @@ def used_kwargs(method, kwargs, ignore_keywords=False):
...
+    Maintains ElasticSearch connection. Picks the ES configuration from 
+    environment variables.
+    '''
+    connection = None
+    def get_connection(self):
+        # check if connection is created or still alive
+        if self.connection is None or not self.connection.ping():
+            self.connection = Elasticsearch(variables['ES_HOST'],
+                http_auth=(variables['ES_USER'], variables['ES_PASS']))
+        return self.connection
```
sanand0, line 880: Can we remove all classes here? We're not really re-using them. Instead a cached function like this could work:

```
es_conf = {}
def get_connection():
    if 'connection' not in es_conf:
        es_conf['connection'] = ...
     return es_conf['connection']
```

gramex/config.py
```diff
@@ -840,3 +844,78 @@ def used_kwargs(method, kwargs, ignore_keywords=False):
...
+
+
+def log_to_es():
+    '''
+    A thread which picks the logs from log() via a queue and stores in ElasticSearch.
+    todo: do a bulk indexing.
+    '''
+    while True:
+        time.sleep(0.1)
```
sanand0, line 889: OPTIONAL: Can we use the async features? We can assume Python 3.7+. We can avoid threads, and run this in an event loop instead. Very very low priority. (If this is not achieved, I'll still merge. If it takes more than 2 hrs, drop it.)

gramex/config.py
```diff
@@ -840,3 +844,78 @@ def used_kwargs(method, kwargs, ignore_keywords=False):
...
+        try:
+            log_doc = es_q.get()
+            es = ElDb().get_connection()
+            es.index(index=variables['ES_INDEX'], body=log_doc)
+        except Exception:
```
sanand0, line 894: This catches a generic exception and ignores it, except for writing the error. Avoid this try-except block.

gramex/config.py
```diff
@@ -840,3 +844,78 @@ def used_kwargs(method, kwargs, ignore_keywords=False):
...
+    '''
+    A thread which picks the logs from log() via a queue and stores in ElasticSearch.
+    todo: do a bulk indexing.
+    '''
+    while True:
+        time.sleep(0.1)
+        try:
+            log_doc = es_q.get()
+            es = ElDb().get_connection()
+            es.index(index=variables['ES_INDEX'], body=log_doc)
```
sanand0, line 893: What happens if ES_INDEX is not present?

gramex/config.py
```diff
@@ -840,3 +844,78 @@ def used_kwargs(method, kwargs, ignore_keywords=False):
...
+        except Exception:
+            app_log.error('unable to store in ES')
+
+if esl_conf['enabled']:
```
sanand0, line 897: If `eslog:` is specified, assume that we want it enabled.

Pass parameters like the index and any other configurations through the gramex.yaml.

gramex/services/__init__.py
```diff
@@ -190,6 +190,8 @@ def callback():
                 def check_exit():
                     if exit[0] is True:
                         shutdown()
+                        # gramex.config.c.terminate()
```
sanand0, line 193: Remove comments

sanand0:
After these, the next step is writing test cases and adding documentation (in the Gramex Guide repo).

PR #275, opened 2020-08-11
BLD: Update Dockerfile

jaidevd:
Add a "make docker" command.


sanand0:
I've added commit ae3f2edc which:

- Uses the low-level Docker API to stream the output -- so I can see the docker build executing live
- Uses `gramex.cache.open(rel=True)` to reduce the lines of code a bit
- Uses maintainer and description labels from release.json

If you're OK, I'll merge.

jaidevd:
LGTM, but can't we stream w/o the low-level API? If we can't, that means we're not using the return value of the [`client.images.build`] (https://docker-py.readthedocs.io/en/stable/images.html#docker.models.images.ImageCollection.build) correctly.

pkg/docker-py3/build.py
```diff
@@ -0,0 +1,14 @@
...
+import io
+
+template = gramex.cache.open('Dockerfile.tmpl', 'template', rel=True)
+release = gramex.cache.open('../../gramex/release.json', rel=True)
+client = docker.from_env()
+streamer = client.api.build(
+    fileobj=io.BytesIO(template.generate(**release)),
+    tag=f'gramener/gramex:{release["version"]}',
+    decode=True,
+)
```
jaidevd, line 12: Please add `rm=True` as an argument here, otherwise daemon creates a lot of intermediate images and doesn't clean them up on a successful build.

sanand0:
Merged via 24c41d09

PR #281, opened 2020-09-04
ElasticSearch logging

sanand0:


gramex/services/__init__.py
```diff
@@ -902,3 +904,38 @@ def test(conf):
...
     # If there are passwords here, they will not be loaded in memory
     conf.pop('auth', None)
+
+
+def gramexlog(conf):
+    '''
+    '''
+    from elasticsearch import Elasticsearch
+    from elasticsearch import helpers
+    default_conf = conf.get('default')
```
sanand0, line 914: This only works for the `default:` app. It won't work for other servers. You'll need one connection for each server.

gramex/__init__.py
```diff
@@ -291,3 +292,51 @@ def shutdown():
...
+    if handler is not None:
+        kwargs.update({key: val[0] for key, val in handler.args.items()})
+    from . import services
+    conf = services.info.gramexlog
+    if conf and 'queue' in conf and 'maxlength' in conf:
+        kwargs['level'] = kwargs.get('level', 'INFO').upper()
+        kwargs['time'] = datetime.datetime.utcnow().strftime('%Y-%m-%d %H:%M:%SZ')
+        kwargs['port'] = app_log_extra['port']
+        idx = kwargs.get('_index', 'default')
+        kwargs['_index'] = conf['conf'].get(idx).get('index')
```
sanand0, line 337: Let's handle these situations:

- `log(_index='nonexistent')` -- raise an Exception. No need to handle
- The `index:` key is not specified in the configuration. Then, default it to the app name
- Allow multiple applications

So this code can go into `log_to_es()`
```python
        idx = kwargs.get('_index', 'default')
        kwargs['_index'] = conf['conf'][idx].get('index', idx)
```

gramex/services/__init__.py
```diff
@@ -902,3 +904,38 @@ def test(conf):
...
+        default_conf.get('host') or None, http_auth=(
+            default_conf.get('user'), default_conf.get('pass')))
+
+    def log_to_es():
+        try:
+            if queue:
+                # Override _id -- a must for bulk indexing in ElasticSearch
+                for item in queue:
+                    item['_id'] = uuid.uuid4()
+                helpers.bulk(connection, queue)
```
sanand0, line 929: Select the connection based on the `_index` in the queue. This means you also need to know which host, user ID, password, etc.

gramex/gramex.yaml
```diff
@@ -216,6 +216,18 @@ log:
 eventlog:
     path: $GRAMEXDATA/logs/events.db          # Location of application event logs.
 
+gramexlog:
+    default:
+        host: $ES_HOST
+        user: $ES_USER
+        pass: $ES_PASS
+        index: gramexlog
+    app:
```
sanand0, line 225: Remove this from `gramex/gramex.yaml` but add it to `tests/gramex.yaml` with a different hostname (e.g. `localhost` vs `127.0.0.1`)

tests/test_eslog.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # Without eslog config, log queue should be empty
+        # Check if the log is written to elastic search
+
+    def test_eslog_not_running(self):
+        # 1. Create a connection to a nonexistent elasticsearch port and send the request
+        # 2. Stop elasticsearch and send a log message
+        # self.check('/log', {'level': 'WARN', 'x': x, 'msg': 'abc'})
+        pass
+
```
sanand0, line 40: Add a test case to send log messages to `default` and to `app` and ensure that they go to the right index.

sanand0 (commented):
Mainly make this robust for multiple servers and add tests.

gramex/__init__.py
```diff
@@ -322,21 +322,24 @@ def log(handler=None, **kwargs):
...
+            kwargs['port'] = app_log_extra['port']
+            if '_app' in kwargs.keys() and kwargs.get('_app', '') not in conf['conf'].keys():
+                raise IndexError('EsLog not configured for application')
+            if len(conf['queue']) < conf['maxlength']:
+                conf['queue'].append(kwargs)
+            else:
+                raise IndexError('Gramex log queue (%s) is too long (max: %s)' % (
+                    len(conf['queue']), conf['maxlength']))
+    except Exception:
+        pass
```
sanand0, line 345: Use `app_log.exception()` to log the exception. Don't `pass` silently

gramex/__init__.py
```diff
@@ -322,21 +322,24 @@ def log(handler=None, **kwargs):
...
-            conf['queue'].append(kwargs)
-        else:
-            raise IndexError('Gramex log queue (%s) is too long (max: %s)' % (
-                len(conf['queue']), conf['maxlength']))
+    try:
+        if conf and 'queue' in conf and 'maxlength' in conf:
+            kwargs['level'] = kwargs.get('level', 'INFO').upper()
+            kwargs['time'] = datetime.datetime.utcnow().strftime('%Y-%m-%d %H:%M:%SZ')
+            kwargs['port'] = app_log_extra['port']
+            if '_app' in kwargs.keys() and kwargs.get('_app', '') not in conf['conf'].keys():
```
sanand0, line 337: This can be

```python
if '_app' in kwargs and kwargs['_app'] not in conf['conf']:
```

gramex/services/__init__.py
```diff
@@ -909,24 +908,31 @@ def test(conf):
 def gramexlog(conf):
     '''
     '''
+    # import uuid
     from elasticsearch import Elasticsearch
-    from elasticsearch import helpers
+    # from elasticsearch import helpers
```
sanand0, line 913: Remove this line -- and any other commented code

sanand0:
Merged via b74a514d

Issue #256, opened 2020-07-07
CaptureHandler crashes on long-running browser instance


mniyas:
CaptureHandler crashes with the following error after running for long duration
`WebSocket is not open: readyState 3 (CLOSED)`

Related bug: https://github.com/puppeteer/puppeteer/issues/4428

Possible fix:
```
 chromecapture.js#132

  // if (typeof browser == 'undefined') 
  //     browser = await puppeteer.launch({args: args})
  browser = await puppeteer.launch({args: args})
  browser.on('disconnected', () => {
    render(q);
  });
```
The commented block is checking for `browser` type and is returning false in 2nd call. I've commented this. Didn't understand the need for this check though.

sanand0:
This looks like it could work. You may need to set `browser = undefined` as well.

It would be good to write a test case so that when `?_test_disconnect=1`  is passed, it disconnects the browser and tests this branch, please.

sanand0:
@mniyas I'd like to include this fix in this month's release. Would you be able to spare time to work on the test case? If not, we'll take it up.

PR #291, opened 2020-09-15
FIX: #256: Relaunch browser on disconnect

mniyas:


mniyas:
@sanand0 Please review. The failed test case is not related to this PR.

sanand0:
Thanks @mniyas -- the test case failure is due to TwitterRESTHandler. I'll explore that later.

I'll review this, test it locally, and revert with changes if required by EOD. Else I'll merge.

tests/test_capturehandler.py
```diff
@@ -318,3 +318,9 @@ def test_pdf_header_footer(self):
         # Text elements are grouped differently in PDFs generated by Window and Linux
         ok_('Header' in layout[0] or 'Header' in layout[1])
         ok_(re.search(r'1\s*/\s*1', layout[-1]))
+
+    def test_browser_disconnect(self):
+        result = self.fetch(self.src, params={
+            'url': self.url, 'file': 'capture', 'ext': 'jpg', '_test_disconnect': 1})
+        self.check_filename(result, 'capture.jpg')
+        self.check_img(result.content)
```
sanand0, line 326: I don't think this really checks if the browser restarts after disconnecting. This saves the file, and THEN disconnects. After that, we're not really testing anything.

We should disconnect, and then request ANOTHER file. That'll prove that it's able to reconnect.

gramex/apps/capture/chromecapture.js
```diff
@@ -226,6 +238,11 @@ async function render(q) {
     }
   }
   await page.close()
+  if (url.parse(q.url, true).query._test_disconnect == 1){
```
sanand0, line 241: I would suggest using just `if (q._test_disconnect) {`

`_test_disconnect` will get passed as part of `q` itself, not the URL encoded. So this code block can just be

```js
if (q._test_disconnect) {
  console.log('Disconnecting Chrome')
  browser.disconnect()
}
```

gramex/apps/capture/chromecapture.js
```diff
@@ -129,7 +140,8 @@ async function render(q) {
     args.push('--proxy-server=' + proxy)
 
   if (typeof browser == 'undefined')
-    browser = await puppeteer.launch({args: args})
+    browser = await browser_setup(args)
```
sanand0, line 143: We need to make another check here. If `browser` is undefined, OR if `browser` is disconnected, we should run this setup.

gramex/apps/capture/chromecapture.js
```diff
@@ -74,6 +76,15 @@ function templatize(input) {
     .join('')
 }
 
+const browser_setup = async (args) => {
```
sanand0, line 79: Since this function is used only once, you can write it inline. No need for a separate function.

gramex/apps/capture/chromecapture.js
```diff
@@ -11,6 +11,8 @@ const path = require('path')
 const tmp = require('tmp')
 const fs = require('fs')
 const _ = require('lodash')
+const url = require('url')
```
sanand0, line 14: I believe this won't be required. In a comment below, I suggest that we just use `q._test_disconnect` instead of `url.parse`

sanand0 (changes requested):
@mniyas I've suggested changes to this PR. The main thing is that the test case doesn't actually test the disconnection behavior. Please take a look. Happy to help if you have any questions. Thanks!

mniyas:
@sanand0 Please note the change in test case

sanand0:
Thanks @mniyas -- I'll review this today

sanand0:
Closed via 1698689

PR #295, opened 2020-09-22
Offline installation on Linux

jaidevd:
This PR adds a `make offline` option which creates a tarball, to be moved to the target machine. It includes an `install.sh` script, which installs Gramex.


pkg/offline/offline-setup.py
```diff
@@ -0,0 +1,17 @@
+import os
```
jaidevd, line 1: This, and the conda template generating scripts can go in the same file.

pkg/offline/install.tmpl
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
jaidevd, line 1: Mention that a conda environment with python=3.7 and a pip that supports `--use-feature=2020-resolver` must be present on the target machine.

Makefile
```diff
@@ -74,6 +74,12 @@ conda:
 	python pkg/conda/conda-setup.py
 	conda build -c conda-forge pkg/conda/
 
+offline:
+	python pkg/conda/conda-setup.py
+	python pkg/offline/offline-setup.py
+	bash pkg/offline/build.sh
+	@echo "Move ./gramex-offline.tar.bz2 to the target machine, unpack it, and run the 'install.sh' script from the same folder."
```
jaidevd, line 81: Put this in a README maybe?

.gitignore
```diff
@@ -152,3 +152,14 @@ docs/_build
...
+# Files produced by make offline
+*.tar.bz2
+pkg/offline/GRAMEXAPPS/
+pkg/offline/GRAMEXDATA/
+pkg/offline/_conda_offline_pkg/
+pkg/offline/_conda_offline_recipe/
+pkg/offline/conda/
+pkg/offline/pip/
+pkg/offline/build.sh
+pkg/offline/install.sh
```
jaidevd, line 165: Only files that _don't_ need to be ignored are the templates.

sanand0:
@jaidevd is this is automating what we have at https://gramener.com/gramex/guide/install/#offline-install? If yes, it sounds like too much code. If it's doing more, can we discuss?

jaidevd:
@sanand0 Yes. The only other thing it's doing is making sure that the effect of `gramex setup --all` is also present on the target machine. Please see https://github.com/gramener/gramex/blob/3f11c8a27af3980de60aa82a40e0b5f595f5226a/pkg/offline/build.tmpl

This is the largest file, can you suggest what can be trimmed?

sanand0:
I was thinking of an alternate approach:

```bash
wget https://repo.anaconda.com/archive/Anaconda2-5.2.0-Linux-x86_64.sh
bash Anaconda2-5.2.0-Linux-x86_64.sh -b -p /tmp/gramex
/tmp/gramex/bin/conda install -y -c conda-forge -c gramener gramex
tar -cvzf /tmp/gramex.tar.gz /tmp/gramex/
```

Wouldn't this do the entire job?

jaidevd:
This basically works, but if we want to use this as a basis to create a script which will generate the tarball; then

1. The Anaconda download URL will keep changing.
2. The `gramex` executable has a hashbang which will have to be respected on the client machine.
3. The resulting tarball is almost a gigabyte.

Other than that, I don't see a problem.

sanand0:
Next steps:

- @jaidevd to test this with Miniconda instead of Anaconda, and share the final lines to copy-paste
- @sanand0 to test this on Windows, and add this to the guide

jaidevd:
@sanand0 This is what works:
```bash
wget https://repo.anaconda.com/miniconda/Miniconda3-py37_4.8.3-Linux-x86_64.sh -O /tmp/miniconda.sh
bash /tmp/miniconda.sh -b -p /tmp/gramex-offline
/tmp/gramex-offline/bin/conda create -n gramex python=3.7 -y
/tmp/gramex-offline/bin/conda install -y 'pip>=20.2'
/tmp/gramex-offline/bin/conda install -y -c conda-forge -c gramener gramex
tar -jvcf /tmp/gramex-offline.tar.bz2 /tmp/gramex-offline/
```

jaidevd:
@kamleshdjango, checkout this branch, and run `make offline` from the gramex directory. It will create an archive named `gramex-offline.tar.bz2`.

Move the created file to the target machine, unpack it, and run the 'install.sh' script from the same folder.

Please try this out - this will help you avoid editing the hashbangs in the conda and gramex scripts.

sanand0:
Closed as outdated

Issue #303, opened 2020-09-30
Extendability - ModelHandler


bkamapantula:
- accept import paths (scikit, tensorflow, keras) in YAML similar to path arg currently for files
- complete an end-to-end ML task with ModelHandler: add, train, test, predict, download
- add a demo for this
- include a DL model

jaidevd:
Thanks, @bkamapantula 
As a next step, @sanand0 , could we find a way to let Gramex users register their own callbacks for `gramex.cache.open`?

sanand0:
Agreed. Adding to the 1 Nov release.

sanand0:
@jaidevd The master branch lets you register custom callbacks. For example:

```python
        gramex.cache.open_callback['shp'] = geopandas.read_file     # Register
        prs = gramex.cache.open('my.shp', layer='countries')        # Open with method
```

[See docs](https://learn.gramener.com/gramex/gramex.html#gramex.cache.open)

Now, what do we do to accept TensorFlow or Keras models?

jaidevd:
Keras models are loaded as follows:
```python
>>> from tensorflow.keras.models import load_model
>>> model = load_model('mymodel.h5', **kwargs)
```
`*.h5` is already a callback. So I'd suggest that we find a way to invoke the Keras callback _only_ when it's within a `ModelHandler` scope. After that, we just have to make sure that Keras models are invoked via Tensorflow's [`scikit-learn` wrappers](https://www.tensorflow.org/api_docs/python/tf/keras/wrappers/scikit_learn)

That should do it.

jaidevd:
BTW, in a recent release, scikit-learn estimators and pipelines come with an [HTML representation](https://scikit-learn.org/stable/auto_examples/compose/plot_column_transformer_mixed_types.html#html-representation-of-pipeline).

It would be really cool if we had a feature such that when a ModelHandler is specified in gramex.yaml, it includes by default an endpoint which shows the HTML diagrams of all models under it.

Suppose:
```yaml
url:
  modelhandler:
    pattern: /$YAMLURL/model/(.*?)/(.*?)
    handler: ModelHandler
    kwargs:
      path: $YAMLPATH  # The local directory to store model files/training data etc.
```

Then, something like `/$YAMLURL/model/_prettify` shows the diagrams.

pratapvardhan:
[...]

PR #310, opened 2020-10-13
ENH: Get predictions from Keras models

jaidevd:
This PR allows ModelHandler to accept a Keras model, and draw
predictions from it. See #303 

@sanand0, @bkamapantula and I have been exploring how to make Keras models work with `ModelHandler`. [Here](https://github.com/jaidevd/chart-ann/blob/master/chart_annotation.py#L37) is the subclass we wrote to be able to POST image bytes to the handler and get a predicted label from a pretrained model. As you can see, we barely reused anything in ModelHandler, since almost every method that matters is too tightly coupled with sklearn.

On this branch, we can make the [same app](https://github.com/jaidevd/chart-ann) work with Keras without any subclassing. The changes are still very hacky, though. Need your inputs on how to make `ModelHandler` less sklearn dependent.


gramex/handlers/modelhandler.py
```diff
@@ -6,16 +6,44 @@
...
+
+        Parameters
+        ----------
+        path : str
+            Path to a Keras model.
+        """
+        model = _model_cache.get(path, False)
+        if not model:
+            _model_cache[path] = gramex.cache.open(path)
+        return _model_cache[path]
```
pratapvardhan, line 30: @jaidevd -- Since you're using `gramex.cache` you can probably get rid of `_model_cache` another cache check?

gramex/handlers/modelhandler.py
```diff
@@ -6,16 +6,44 @@
...
+
     @classmethod
     def setup(cls, path, **kwargs):
+        if os.path.splitext(path)[-1] == '.h5':
+            from tensorflow.keras.models import load_model
+            gramex.cache.open_callback['h5'] = load_model
```
pratapvardhan, line 36: @jaidevd -- this would override `.h5` callback across the app. Which means, I can't use `pd.read_hdf`. Instead use 

```python
from tensorflow.keras.models import load_model
cls._model_loader = lambda path: gramex.cache.open(path, load_model)
```
perhaps?

gramex/handlers/modelhandler.py
```diff
@@ -6,16 +6,44 @@
...
+            cls.model_type = "keras"
+            cls._model_loader = cls._load_keras_model
+        else:
+            cls.model_type = "sklearn"
+            cls._model_loader = staticmethod(lambda x: gramex.cache.open(x, gramex.ml.load))
```
pratapvardhan, line 41: @jaidevd -- should `gramex.ml.load(..., engine='keras')` handle both keras ans sklearn instead here?

jaidevd: `gramex.ml.load` is just `joblib.load` - should we add an explicit `gramex.ml.load` that does this (takes the `engine` kwarg and decides what to do)?

jaidevd:
@pratapvardhan Done... with `gramex.cache.open` and `gramex.ml.load`, much smaller change now.

sanand0 (changes requested):

jaidevd:
@sanand0 This is done. External `prepare` methods can be added now. The following config is valid:

```yaml
modelhandler:
    pattern: /$YAMLURL/predictChart/
    handler: ModelHandler
    kwargs:
      path: $YAMLPATH/models/four-class.h5
      prepare: chart_annotation.prepare  # this transforms the binary stream in the request to an np.ndarray
      engine: keras
      xsrf_cookies: false
```

The caveat with Keras models is that we only support image models, and the responsibility of transforming the request body into numpy / tensorflow arrays is on the `prepare` method provided by user. ModelHandler by itself won't do any preprocessing.

We can, however, provide utils that do this in `gramex.ml`, for e.g:

```
from io import BytesIO
from skimage.io import imread

def img_from_bytes(handler):
       handler.request_body['image'] = imread(BytesIO(handler.request.body))
```

which users can use as:
```yaml
modelhandler:
    pattern: /$YAMLURL/predictChart/
    handler: ModelHandler
    kwargs:
      path: $YAMLPATH/models/four-class.h5
      prepare: gramex.ml.img_from_bytes
      engine: keras
      xsrf_cookies: false
```

sanand0:
Closed in favor of a full rewrite of (or replacement for) ModelHandler

Issue #304, opened 2020-10-01
Convert any function into a request handler


jaidevd:
Add a decorator which converts any function to work with a `handler` object.

Suppose we have:
```python
@make_handlerfunc('GET')
def add(x, y, z=42):
   return x + y + z

@make_handlerfunc('POST')
def greet(name):
    return f"Hello, {name}!"
```

then, this would become equivalent to:
```python
def wrapped_func(handler):
    args = handler.path_args
    if handler.request.method == "GET":
        kwargs = handler.query_params  # or whatever
    elif handler.request.method == 'POST':
        kwargs = handler.request_body_args
    return original_func(*args, **kwargs)
```

That way, I can add `add` or `greet` directly in my application config:
```yaml
foo:
  pattern: /$YAMLURL/(.*)/(.*)
  handler: FunctionHandler
  kwargs:
    # Note that this is `add` as defined above.
    # By itself, it won't work, but the `make_handlerfunc` wrapper makes it work.
    function: mymodule.add
```

For now, we can enforce that only functions which return a JSON-able object will work with this. Let's also generally assume that it's the user's responsibility to make the wrapped function otherwise RESTful.

We will only infer whatever we can from function signatures (args, kwargs) and type hints. We will also force `json.dumps` the output.

sanand0:
Agreed. Let's do this. Would you like to implement this, @jaidevd?

jaidevd:
Yes, @sanand0 I'll send a PR on Wednesday.

jaidevd:
@sanand0 , this is a minimal example:
```python
class MockHandler(object):
    def __init__(self, path_args=None, urlparams=None, body=None,
                 method='GET'):
        if not path_args:
            path_args = []
        self.path_args = path_args
        if not urlparams:
            urlparams = {}
        self.urlparams = urlparams
        self.body = body
        self.method = method


def decorator(func):
    def wrapper(handler):
        pargs = handler.path_args
        if handler.method == 'GET':
            kwargs = handler.urlparams
        elif handler.method == 'POST':
            kwargs = handler.body
        return func(*args, **kwargs)
    return wrapper


@decorator
def _add(a, b):
    return a + b


if __name__ == '__main__':
    handler = MockHandler(path_args=[1, 2])
    # Now this works with a handler:
    print(_add(handler))  # 3
```

Now we just generalize the decorator until it works with most cases.

sanand0:
[...]

PR #311, opened 2020-10-14
Convert functions to accept RequestHandlers

jaidevd:
See #304 
This PR adds a decorator: `gramex.handlers.functionhandler.add_handler` which wraps any callable to:

1. get it's args and kwargs from handler.path_args and handler.request.body
2. return a JSON object
3. enforce any type annotations found in the callable

ToDo:
- [x] ~~Not exhaustively tested for POST / DELETE~~


gramex/handlers/functionhandler.py
```diff
@@ -1,13 +1,49 @@
...
+    func : callable
+        function to be wrapped.
+    """
+    sig = signature(func)
+
+    @wraps(func)
+    def wrapper(handler):
+        args = handler.path_args
+        if handler.request.method == 'GET':
+            kwargs = {k: v[0] for k, v in handler.args.items()}
```
jaidevd, line 31: `handler.get_arg` does this.

jaidevd: Test all HTTP methods.
Test all native types including `None`.
Test what happens with default named arguments.
`?x=1&x=2&x=3` should be doable. Perform the typecasting from the hints.

pratapvardhan, line 31: Perhaps `{k: v if len(v) > 1 else v[0] for k, v in handler.args.items()}` to handle `?x=1&x=2&y=3`?

gramex/handlers/functionhandler.py
```diff
@@ -1,13 +1,49 @@
...
+            kwargs = handler.body
+        hints = get_type_hints(func)
+        for arg, argtype in hints.items():
+            param = sig.parameters.get(arg, False)
+            if param:
+                if param.kind == param.VAR_POSITIONAL:
+                    args = [argtype(k) for k in args]
+                else:
+                    kwargs[arg] = argtype(kwargs[arg])
+        return json.dumps(func(*args, **kwargs))
```
pratapvardhan, line 42: Thoughts on using pydantic? See this example https://pydantic-docs.helpmanual.io/usage/validation_decorator/

gramex/handlers/functionhandler.py
jaidevd: @sanand0 This ^ works. `handler.get_arg('x')` only gets the last one.

jaidevd:
@sanand0 items from the last review are done. Some notes on usage are in the [docstring for `add_handler`](https://github.com/gramener/gramex/blob/5c57e1106e41d93a2583d728757502df33d7a0c9/gramex/handlers/functionhandler.py#L41). Please take a look.

gramex/handlers/functionhandler.py
jaidevd: @pratapvardhan this is great... I can offload all the typechecking / casting off to `pydantic.validate_arguments`.
Type validation should be on by default, and in that case, pydantic should do it. But users should also be able to turn it off if required, like:
```python
@add_handler(validate=False)
def slow_sum(*numbers: float, delay: int):
   s = 0
   for i in numbers:
       time.sleep(delay)
       s += 0
       yield s
```
This ^ snippet won't complain when `numbers` is not floats, say.

@sanand0 Thoughts?

sanand0: No inputs -- and we can defer this.

sanand0:
- [ ] Move `add_handler` as `gramex.transforms.handler`
- [ ] Fix typos in docstring
    - `greet.,py` -> `greet.py`
    - `funtion` -> `function`
    - `andd` -> `and`
    - Please check for others
- [ ] Change usage to:

```yaml
pattern: /$YAMLURL/greet
handler: Function
kwargs:
    function: greet.birthday
```

- [ ] This test case should

```python
def total_list(items: List[float], start: float) -> float:
    return sum(items) + start
```

- [ ] In POST and PUT requests, URL parameters and path arguments should be used. Add test cases

sanand0:
These scenarios don't yet work.

```yaml
url:
  power:
    pattern: /power
    handler: FunctionHandler
    kwargs:
      function: alpha.power
      kwargs:
        x: 3

  message:
    pattern: /message
    handler: FunctionHandler
    kwargs:
      function: alpha.messages
```

```python
@handler
def power(x: int, y:float) -> float:
    return y ** x


@handler
def messages(s: list):
    for msg in s:
        yield msg
```

jaidevd:
@sanand0 Added these two cases, and tests for them.

sanand0:
I made these changes:

- [x] Make all handler names, URL patterns and function names identical
- [x] Move `@handler` methods from `test_function.py` to `functionutils.py`. This is to allow manual testing. Specifically, if `cd tests` and run `gramex` to visit `localhost:9999`, Gramex can't import `tests.test_function` because it has a line `from . import TestGramex`.
- [x] Split tests into a `test_functionhandler.TestWrapper` class so we can run them independently
- [x] If `?x=1&x=2` are passed to `func(x: int)`, then the **last** parameter is used, not the first
- [x] POST body is processed as JSON only if `Content-Type: application/json`. This is consistent with FormHandler. It also retains the URL query parameters apart from the body
- [x] If we call `/func/total/10/20?items=30&items=40` it will add 10+20+30+40=100. THis required a rewrite

sanand0:
Note: Checks failed because Travis couldn't access the Microsoft R repository. Even [master fails](https://travis-ci.com/github/gramener/gramex/builds/191618600) despite the same commit passing earlier. So I've gone ahead and merged.

Issue #309, opened 2020-10-12
Extend gramex.cache.open to read named ranges and workbook tables


sanand0:
**Is your feature request related to a problem? Please describe.**
A single Excel sheet may contain multiple [tables](https://support.microsoft.com/en-us/office/overview-of-excel-tables-7ab0bb7d-3a9e-4b56-a3c9-6c94334e492c) or [named ranges](https://support.microsoft.com/en-us/office/define-and-use-names-in-formulas-4d0f13ac-53b7-422e-afd2-abd7ff379c64). We need to read from these as well. Pandas does not directly support this but openpyxl does.

**Describe the solution you'd like**

Access specific cells via:

```python
gramex.cache.open('file.xlsx', 'xl', sheet_name='Sheet1', range='A1:C10')
gramex.cache.open('file.xlsx', 'xl', sheet_name='Sheet1', name='My Named Range')
gramex.cache.open('file.xlsx', 'xl', sheet_name='Sheet1', table='My Table')
```

sanand0:
This will be important for SlideSense. I'll plan this for 1.65

PR #318, opened 2020-10-27
Extend gramex.cache.open to read named ranges and workbook tables #309

sanand0:


gramex/cache.py
```diff
@@ -128,6 +128,45 @@ def _template(path, **kwargs):
...
+        range_str = wb.defined_names.get(name, sheetid).attr_text.split('!')[1].replace('$', '')
+        get_rows(sheet_name, range_str)
+
+    if table is not None:
+        get_rows(sheet_name, wb[sheet_name].tables[table].ref)
+
+    header_val = None
+    if header is not None:
+        header_val = list(data_rows[header])
+        del data_rows[header]
```
pratapvardhan, line 164: @vinothsm -- I suspect if this handles `header can be list of int` scenario. If you're handling only int and None, you could potential simplify with .. ?

```python
header = None if header is None else list(data_rows.pop(header))
df = pd.DataFrame(data_rows, columns=header)
```
without needing to `del` and intermediate `header_val`.

gramex/cache.py
```diff
@@ -128,6 +128,45 @@ def _template(path, **kwargs):
...
 
 
+def excel_reader(path, sheet_name='Sheet1', range=None, name=None, table=None, header=0):
+    # import openpyxl
+    # Read the file using openpyxl
+    # If range (e.g. A10:C30) is specified, return cells in that range
+
+    # If name is specified, return cells in that named range using wb.defined_names
+    # If table is specified, return cells in that named table using ws.tables
+    # Convert into a DataFrame -- using header if specified. header can be int, list of int, None
```
pratapvardhan, line 138: What would happen if you use both `range` and `name` and or `table`? If it's only intended to handle exclusive cases, `if, else if` control logic might be appropriate here no?

gramex/cache.py
vinothsm: @pratapvardhan  Thanks for reviewing. `header` is int only and agree with your suggestion to simplify like above.

gramex/cache.py
vinothsm: @pratapvardhan We are yet to decide the behavior in such scenarios. @sanand0 Could you please provide your thoughts?

sanand0: table overrides name overrides range. I've implemented this

gramex/cache.py
sanand0: Have implemented list of int as a possibility.

PR #328, opened 2020-11-04
Allow ModelHandler URLs to be relative to YAMLPATH

jaidevd:
`ModelHandler.path` is the path to the model - and should have nothing to do with the URL / path to the training dataset. Currently, we assume that models and data are in the same directory. This PR stops assuming that the training dataset URLs are relative to the `path`
kwarg of ModelHandler. Instead, makes them relative to $YAMLPATH.


sanand0:
Closed in favor of a full rewrite of (or replacement for) ModelHandler

PR #344, opened 2020-12-03
Redis bugs

mniyas:
Fixed 3 issues.

1. Redis Session store is not taking path with `=`. This was fixed for Redis cache
2. Redis Cache uses pickled keys. However, in case if the Redis has unpickled keys retrieving keys fails.
3. Some managed Redis services (eg: Azure) won't allow changing config from client-side. Hence setting the default Redis cache size as `None`. Now, the user can remove the `size` parameter in `gramex.yaml` and set it at the server.


sanand0:
Thanks @mniyas -- this looks good. Any chance you could add test cases for these, please?

mniyas:
@sanand0 Added test case with additional change.

sanand0:
- [ ] Please commit to the mniyas-redis branch
- [ ] Add the `.decode('utf-8')` to wherever required in cache.py
- [ ] Ensure that the test case(s) covers all the lines changed
- [ ] @sanand0 to document that the default redis cache size is 0, and can be overriden

mniyas:
@sanand0 I have modified the test case. Gramex cache keys are tuples and need not be decoded. Redis keys can be anything and can be ignored if they are not set by Gramex.

sanand0:
Closed in favor of #349

PR #349, opened 2020-12-16
Redis bugs

sanand0:


sanand0:
- [ ] @sanand0 Write test case to check -- when `size: 0` is specified, the Redis maxmemory is never set. (Required for Azure)
- [ ] @sanand0 to document that the default redis cache size can be overridden as 0

sanand0:
Merged via 28103e92

PR #357, opened 2021-01-06
FIX: Viewing an MLHandler template drops the target col

### LINKED ISSUE #361: MLHandler should show data alongwith predictions and score
When `action=score` or `action=predict`, MLHandler returns only the predicted column (a JSON array) or the score. It should return the whole dataframe which was POSTed, predicted column as a new column. The score should be sent separately as a different key.

Currently:
```
curl -X POST -F 'file=@data.csv' /mlhandler?_action=predict
```
returns just `[0, 1, 0, 1, 0, 1, ....]`

It should instead return the original dataset with the new column appended.

jaidevd:
Rendering any template in MLHandler accidentally drops the target
column. Fixing this. Closes #361

### DIFF
--- a/gramex/handlers/mlhandler.py
+++ b/gramex/handlers/mlhandler.py
@@ -5,14 +5,13 @@ import os
 from urllib.parse import parse_qs
 
 import gramex
-from gramex.config import app_log
+from gramex.config import app_log, CustomJSONEncoder
 from gramex import data as gdata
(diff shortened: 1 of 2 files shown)


jaidevd:
I'll fix #361 in this PR itself.

jaidevd:
@sanand0 This is ready.

PR #364, opened 2021-01-30
REQ: Restrict nodejs to LTS versions

jaidevd:
The latest nodejs version available under conda-forge is not stable,
which causes this error:
https://stackoverflow.com/questions/64612707/node-sass-does-not-yet-support-your-current-environment-windows-64-bit-with-uns

### DIFF
--- a/pkg/conda/template.meta.yaml
+++ b/pkg/conda/template.meta.yaml
@@ -41,7 +41,7 @@ requirements:
   run:
     - python          # REQ: conda: Base Python
     - pip             # REQ: conda: Set up pip packages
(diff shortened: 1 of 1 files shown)


bhatsandeep:
Acknowledged!

Will triage, prioritize this issue!

sanand0:
The error we get is: `Error: Node Sass does not yet support your current environment: Windows 64-bit with Unsupported runtime (88)`

For anyone who wants to fix this manually on Gramex 1.68 or less, go to the gramex conda environment. Then run:

```bash
conda install -c conda-forge nodejs==14.15.1`
gramex setup ui
```

PR #365, opened 2021-02-01
Initialize tables on first load in forms 

bkamapantula:
## Problem
forms.db and analytics, forms tables need to be created before forms app can be used.

## Solution
- introduce [formhandler](https://learn.gramener.com/guide/formhandler/#formhandler-columns) columns-based table creation for `analytics` and `forms` tables
  - downstream impact: update data-source for `.formhandler()` in form.html
- disable mouse events while previewing the form at `/form/id`
  -  add a prominent way to View form in addition to a link in navbar 

Without this, forms app wouldn't work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
+# ignore todo files
 TODO
+*.todo
 
 # Ignore auto-generated documentation
 docs/gramex*.rst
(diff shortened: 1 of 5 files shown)


bkamapantula:
@sanand0 updated the PR, kindly review

PR #370, opened 2021-02-06
BLD: Use local source for conda; fix miniconda version for Docker

jaidevd:
1. Gramex >=1.67 requires git lfs on the build machine. Specifying git sources in the conda recipe breaks it. Use local path for the source instead.
2. In the Dockerfile, pin miniconda3 version to the latest one that has Python 3.7 in the base env.


jaidevd:
@sanand0 bumping this up. This allows conda builds without the build machine having git-lfs.

Issue #367, opened 2021-02-02
UI Factory - Forms - Feature, improvement requests


bkamapantula:
**Is your feature request related to a problem? Please describe.**
Improve Forms builder from a usability standpoint.

**Describe the solution you'd like**

Desired features are described below.

#### Copy or Clone fields
- [ ] copy or clone already added fields (without clicking the add modal again). consider a few fields (email, text, password) are added by a user via modal prompt on Add option click. If the user wants to create another text input, they should be able to copy the created field and modify its properties.

#### Render form from JSON
- [ ] render form from JSON (instead of HTML). Currently the form in `.user-form` is rendered by a saved HTML string. We also save its corresponding JSON configuration. Use it instead to render the form.

#### Name attribute editing
- [ ] Clicking a field brings up its attributes on the right side. "name" attribute behavior should be as follows: 
  - [ ] make it a slug of its label (as user types the label) with an option for the user to change it
  - [ ] user should also be able to edit the name field directly. These edits should be prioritized over slugifying the label.

#### Custom HTML field
- [ ] custom field for HTML. Users should be able to add arbitrary HTML in the custom field.

#### Fork a form
- [ ] Users should be able to fork an existing form.
  - [ ] details: add a fork option wherever there's a Remove option. duplicate database entry and the thumbnail

#### Components

- [ ] 1-10 rating scale - https://stackoverflow.com/a/21231958 with an option to configure the icon
- [ ] password component to have view/hide toggle. Current password field has basic behavior.
- [ ] select, https://codepen.io/lathaoro/pen/qBZMjVK
- [ ] dropdowns, with support for `md`, `lg`, `sm` classes

### Bugs, improvements
- [ ] make background color stick (push to config column) on `path/to/forms/form/id` page. Preview page (for form submission) should retain the background color.
- [ ] it's unclear where to click to Submit the form on `path/to/forms/form/id`. make it prominent, next to Edit field?
- [ ] viewing responses is broken (500 internal server error) if the limit is off (lot more than the records). This leaves a bad end user experience.

**Describe alternatives you've considered**
NA

**Additional context**
NA

bkamapantula:
- bug 2 addressed in https://github.com/gramener/gramex/pull/365
- bug 4 addressed in bk-issue367-b4 branch,

bkamapantula:
- bug 7 addressed in sa-form-components branch

bkamapantula:
[...]

PR #371, opened 2021-02-09
Forms - Select element - Set option as selected based on select value

bkamapantula:
fixes bug 9 in #367


bkamapantula:
this is available via https://github.com/gramener/gramex/pull/388

PR #380, opened 2021-03-03
PKG: Conda - Don't compile untracked .py files from local sources

jaidevd:
Since we started building conda packages from local sources instead of
PyPI (see 13ca54f), all .py files under
the source path get compiled, even the ones in apps/*/node_modules,
which inflates the conda package. This commit enforces a make clean
before make conda.


sanand0:
To be closed in favor of Guide instructions.

PR #388, opened 2021-03-29
UI Factory refactor

bkamapantula:
rename `apps.forms` to `apps.uifactory`

## Added

- introduces a simpler UI (2 columns instead of 3 columns).
- users will be able to fork a form from Gallery page or Form view page.
- uses `component.js` to create custom components
- each component is now updated to the following structure
  - snippets/component/g-componentname.js
- default configuration is from `fields.js`
  -  snippets.json dependency is removed

## Updated
- screenshots are updated on Form updation

## Bugs
- merging of rows to view Form responses had a bug, that's fixed now.

## Removed
- dragula dependency

## TODO
  - subsequently, setup.js should be removed as snippets.json isn't used anymore.


bkamapantula:
Hi @sanand0 this is ready for review

bkamapantula:
- components are renamed from `g-*` to `bs4-*`
- uses uifactory.js from the master branch, a [commit old](https://github.com/gramener/uifactory/blob/7b042904c92abf6f28684fe6e01ed97aac12b659/uifactory.js)

sanand0:
- [ ] If the /create page is updated, don't navigate away from the page without confirmation. Else people will lose their work
- [ ] Button -- if I set borderRadiusPosition I can't disable it
- [ ] Button -- check if borderRadiusSize works
- [ ] For Button iconType, add links to fa-* and bi-* icons in the help
- [ ] Checkbox, Radio -- How to add values with a comma?
- [ ] Checkbox, Radio -- How to specify "yes" just for the first option? Or just the second?
- [ ] Custom HTML does not need a name
- [ ] Custom HTML -- if the label is empty, don't render it
- [ ] Radio -- does not need placeholder
- [ ] Range -- does not have help
- [ ] Select -- does not have help
- [ ] Select -- scrolls beyond screen. I can't see every option AND the select output at the same time
- [ ] Select -- "Select options" -- wasn't clear it is the comma-separated values
- [ ] Select -- "actions-box" -- setting it to true turns it back to false if we re-pick options. Besides, I've no idea what it is
- [ ] Select -- please review other options

bkamapantula:
`bs4-select` now uses the native `select` element instead of bootstrap-select as the latter's behavior isn't consistently implemented to operate as a component as an attribute for another component.

bkamapantula:
bootstrap-select library field is reintroduced as `bs4-multiselect` which defaults to multiple value selection while `bs4-select` is for native HTML select component.

PR #401, opened 2021-04-21
ENH: Support sklearn metrics in MLHandler  

prakrutisingh24:
MLHandler now supports any scoring metric in Scikit-learn.
Metrics can be specified by users using the `_metric` argument, as follows:
```bash
curl -X POST -F "file=@test.csv" 'http://localhost:9988/?_action=score&_metric=f1'
```
Any metrics listed [here](https://scikit-learn.org/stable/modules/model_evaluation.html) are supported.


jaidevd:
Thanks a lot @prakrutisingh24, this looks great!

I'd suggest just a few changes on the code to make this more usable.

gramex/handlers/mlhandler.py
```diff
@@ -148,7 +158,7 @@ def set_opt(cls, key, value):
             transform[key] = value
             cls.config_store.dump('transform', transform)
             cls.config_store.update['transform'] = transform
-        elif key in ('class', 'params'):
+        elif key in ('class', 'params', 'score_pref'):
             model = cls.config_store.load('model', {})
```
jaidevd, line 162: @prakrutisingh24 All of this code in the setup function is running only once at the beginning, i.e. when the model is being _trained_. Scoring the model doesn't make sense here. Scoring actually happens only later when the user sends a POST request. Therefore, all of our scoring code should happen in that function. We can remove all this code.

gramex/handlers/mlhandler.py
```diff
@@ -261,14 +271,56 @@ def _transform(self, data, **kwargs):
...
+                return r2_score(target, predict)
+
+            else:
+                return self.model.score(data,target)
+
+        except ValueError:
+            print('ValueError occured. Please check if the scoring metric is appropriate for the given data!')
+
+
+
```
jaidevd, line 313: Please remove all the print statements here.

jaidevd: Also, the way this function is written, we are restricting the user to choosing between the scoring methods imported at the top of the file.

The intention is to let the user select **any** scoring method, and if it's not applicable, raise an error. We can take any string specified by the user, and make a scorer function out of it: [see this function](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.get_scorer.html). It can be used as follows:

```python
from sklearn.metrics import get_scorer
user_defined_metric = 'accuracy'  # or roc_auc_score, r2, etc.
scorer = get_scorer(user_defined_metric)
accuracy = scorer(model, data, target)
```

gramex/handlers/mlhandler.py
```diff
@@ -261,14 +271,56 @@ def _transform(self, data, **kwargs):
...
             data = self._parse_data(False)
         data = self._transform(data, drop_duplicates=False)
         self.model = cache.open(self.model_path, joblib.load)
         try:
             target = data.pop(score_col)
-            return self.model.score(data, target)
+            s = self._chooseScore(data, target)
+            print('s: ',s)
+            return s
         except KeyError:
```
jaidevd, line 324: Here, the user will be sending the scoring metric as a request parameter, named `_metric`. So we can use the snippet above to change this.

jaidevd (changes requested):

gramex/handlers/mlhandler.py
```diff
@@ -103,12 +104,13 @@ def setup(cls, data=None, model={}, config_dir='', **kwargs):
 
         cls.set_opt('class', model.get('class'))
         cls.set_opt('params', model.get('params', {}))
+        target_col = cls.get_opt('target_col')
```
jaidevd, line 107: @prakrutisingh24 This line is not required.

gramex/handlers/mlhandler.py
```diff
@@ -103,12 +104,13 @@ def setup(cls, data=None, model={}, config_dir='', **kwargs):
...
 
         if op.exists(cls.model_path):  # If the pkl exists, load it
             cls.model = joblib.load(cls.model_path)
         elif data is not None:
             mclass = cls.get_opt('class', model.get('class', False))
-            params = cls.get_opt('params', {})
+            params = cls.get_opt('params', {})
```
jaidevd, line 113: Whitespace at the end of the line, please remove it.

gramex/handlers/mlhandler.py
```diff
@@ -261,14 +263,18 @@ def _transform(self, data, **kwargs):
...
     def _predict(self, data=None, score_col=''):
         if data is None:
             data = self._parse_data(False)
         data = self._transform(data, drop_duplicates=False)
         self.model = cache.open(self.model_path, joblib.load)
         try:
             target = data.pop(score_col)
-            return self.model.score(data, target)
+            metric = self.get_argument('_metric')
+            print('Metric: ',metric)
```
jaidevd, line 275: Please remove the print statement.

gramex/handlers/mlhandler.py
```diff
@@ -261,14 +263,18 @@ def _transform(self, data, **kwargs):
...
+
     def _predict(self, data=None, score_col=''):
         if data is None:
             data = self._parse_data(False)
         data = self._transform(data, drop_duplicates=False)
         self.model = cache.open(self.model_path, joblib.load)
         try:
             target = data.pop(score_col)
-            return self.model.score(data, target)
+            metric = self.get_argument('_metric')
```
jaidevd, line 274: If the user does not provide a metric, we need to fall back on the original model of scoring. Please provide for that case also.

jaidevd (changes requested):
@prakrutisingh24 Almost done, just one minor change is left.

env/pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = /Users/prakruti/opt/anaconda3/bin
+include-system-site-packages = false
+version = 3.8.5
```
jaidevd, line 3: Please remove this file.

gramex/handlers/mlhandler.py
```diff
@@ -62,6 +63,7 @@ class MLHandler(FormHandler):
 
     @classmethod
     def setup(cls, data=None, model={}, config_dir='', **kwargs):
+    	print('hellos')
```
jaidevd, line 66: Please remove this line.

gramex/handlers/mlhandler.py
jaidevd: This line will cause an error if `_metric` is not provided in the URL. The right way to handle this is as follows:

```python
metric = self.get_argument('_metric', False)  # Note that the second argument is returned if the input is not present
if metric:
    scorer = get_scorer(metric)
    return scorer(self.model, data, target)
return self.model.score(data, target)
```

jaidevd (changes requested):

jaidevd:
@prakrutisingh24 You seem to have accidentally added two unnecessary files. The only file affected by this change should be `gramex/handlers/mlhandler.py`

gramex/handlers/mlhandler.py
jaidevd: This is still not resolved.

jaidevd (changes requested):

jaidevd:
@sanand0 Can you take a look at this?

jaidevd (approved):

sanand0:
Thanks @prakrutisingh24

PR #407, opened 2021-05-13
Cross Validation Added

prakrutisingh24:


gramex/handlers/mlhandler.py
```diff
@@ -20,6 +20,9 @@
 from slugify import slugify
 from tornado.gen import coroutine
 from tornado.web import HTTPError
+from sklearn.metrics import get_scorer
+from sklearn.model_selection import cross_val_predict, cross_val_score
+from sklearn.model_selection import cross_val_predict, cross_val_score
```
jaidevd, line 25: This line appears twice.

gramex/handlers/mlhandler.py
```diff
@@ -112,14 +114,26 @@ def setup(cls, data=None, model={}, config_dir='', **kwargs):
             data = cls._filtercols(data)
             data = cls._filterrows(data)
             cls.model = cls._assemble_pipeline(data, mclass=mclass, params=params)
-
             # train the model
             target = data[target_col]
             train = data[[c for c in data if c != target_col]]
+            # cross validation
+            mod = cls.modelFunction()
```
jaidevd, line 121: This is not required. The model is already present as `cls.model`, see line no: 116.

gramex/handlers/mlhandler.py
```diff
@@ -112,14 +114,26 @@ def setup(cls, data=None, model={}, config_dir='', **kwargs):
...
+            CVscore = cross_val_score(mod, train, target)    
+            CV = sum(CVscore)/len(CVscore)
```
jaidevd, line 123: 1. Use `CVscore.mean()`
2. Variable naming has to follow a specified style - do `pip install flake8` and run the `flake8` command against this file, i.e. `flake8 mlhandler.py`, and check the output.

jaidevd:
@prakrutisingh24 In this PR, we are just computing the cross val score when the model is set up for the first time, and simply printing the CV score. What we need is:

- [ ] When _any_ scoring happens, especially if it is triggered by the user through an `?_action=train` or `?_action=score` - the result should be cross validated score
- [ ] By default, cross validation should happen, but if the user wants, they should be able to _turn it off_ through gramex.yaml, and get scores on the entire training dataset
- [ ] sklearn's `cross_val_scoire` contains a parameter, `cv`, that can be set in many different ways. Check the docs for details. MLHandler users should be able to exercise all those options.


Thanks,

gramex/handlers/mlhandler.py
```diff
@@ -40,6 +44,8 @@
     'nums': [],
     'cats': [],
     'target_col': None,
+    'CV': True,
+    'CVargs': []
```
jaidevd, line 48: Let's have a single argument, `cv`, which can take any value, i.e in gramex.yaml, users should be able to write any of the following.

```yaml
cv: false   # disable cross validation
cv: 5        # Use 5 folds
cv:
  cv: 8   # Use 8 folds
  n_jobs: -1  # with an optional other parameter.
```

gramex/handlers/mlhandler.py
jaidevd: This extra line is unnecessary.

gramex/handlers/mlhandler.py
```diff
@@ -112,14 +117,38 @@ def setup(cls, data=None, model={}, config_dir='', **kwargs):
...
             data = cls._filterrows(data)
             cls.model = cls._assemble_pipeline(data, mclass=mclass, params=params)
-
             # train the model
             target = data[target_col]
             train = data[[c for c in data if c != target_col]]
+            # cross validation
+            print('yayyy we are here')
+            cls.CrossValidation(train,target)
+            print('should have printed')
```
jaidevd, line 126: Please remove the prints.

jaidevd (changes requested):

gramex/handlers/mlhandler.py
```diff
@@ -20,6 +20,9 @@
 from slugify import slugify
 from tornado.gen import coroutine
 from tornado.web import HTTPError
+from sklearn.metrics import get_scorer
+from sklearn.model_selection import cross_val_predict, cross_val_score
+from ast import literal_eval
```
jaidevd, line 25: This should not be required.

gramex/handlers/mlhandler.py
```diff
@@ -40,6 +43,7 @@
     'nums': [],
     'cats': [],
     'target_col': None,
+    'CV': True,
```
jaidevd, line 46: Make it lowercase.

gramex/handlers/mlhandler.py
```diff
@@ -112,14 +115,32 @@ def setup(cls, data=None, model={}, config_dir='', **kwargs):
             data = cls._filtercols(data)
             data = cls._filterrows(data)
             cls.model = cls._assemble_pipeline(data, mclass=mclass, params=params)
-
             # train the model
             target = data[target_col]
             train = data[[c for c in data if c != target_col]]
+            # cross validation
+            cls.CrossValidation(train,target)
```
jaidevd, line 122: Make it lowercase.

gramex/handlers/mlhandler.py
```diff
@@ -112,14 +115,32 @@ def setup(cls, data=None, model={}, config_dir='', **kwargs):
...
                 _fit, cls.model, train, target, cls.model_path, cls.name)
         cls.config_store.flush()
 
+    @classmethod
+    def modelFunction(cls, mclass = ''):
+        model_kwargs = cls.config_store.load('model', {})
+        mclass = model_kwargs.get('class', False)
+        if mclass:
+            model = search_modelclass(mclass)(**model_kwargs.get('params', {}))
+            return model
```
jaidevd, line 133: This function is not required.

gramex/handlers/mlhandler.py
```diff
@@ -112,14 +115,32 @@ def setup(cls, data=None, model={}, config_dir='', **kwargs):
...
+    
+    @classmethod
+    def CrossValidation(cls,train,target):
+        mod = cls.modelFunction()
+        CV = cls.get_opt('CV')
+        if CV:
+            CVscore = cross_val_score(mod, X=train, y=target, **literal_eval(json.dumps(CV)))
+            CVavg = sum(CVscore)/len(CVscore)
+            print('Cross Validation Score : ',CVavg)
```
jaidevd, line 142: CV should take place within the train method only.

gramex/handlers/mlhandler.py
jaidevd: We have to support three cases for the `cv` option:

- [ ] If the user sets `cv: false` - then no cross validation happens
- [ ] If the user sets `cv: 4` (or some other integer) pass it straight to `cross_val_score`
- [ ] The default should be `cv: None`, and in this case, the user should not have to write anything in gramex.yaml

jaidevd (changes requested):
The training is happening in [`def _fit`](https://github.com/gramener/gramex/pull/407/files#diff-46e58528a026c29675c0697def48f7bb1616181cdd4924a02d0a68b19cf8191eR53). Cross validation should also happen there.

gramex/handlers/mlhandler.py
jaidevd:
```python
if cv:
    cvscore = cross_val_score(mod, X=train, y=target, cv=cv)
else:
   # Do the usual .fit
```

gramex/handlers/mlhandler.py
```diff
@@ -112,14 +114,23 @@ def setup(cls, data=None, model={}, config_dir='', **kwargs):
             data = cls._filtercols(data)
             data = cls._filterrows(data)
             cls.model = cls._assemble_pipeline(data, mclass=mclass, params=params)
-
             # train the model
             target = data[target_col]
             train = data[[c for c in data if c != target_col]]
+            # cross validation
+            cls.cross_validation(train,target)
```
jaidevd, line 121: Not required here.

gramex/handlers/mlhandler.py
```diff
@@ -112,14 +114,23 @@ def setup(cls, data=None, model={}, config_dir='', **kwargs):
...
                 _fit, cls.model, train, target, cls.model_path, cls.name)
         cls.config_store.flush()
-
+  
+    @classmethod
+    def cross_validation(cls,train,target):
+        cv = cls.get_opt('cv',True)
+        if cv:
+            CVscore = cross_val_score(cls.model.steps[-1][1], X=train, y=target, cv=cv)
+            CVavg = sum(CVscore)/len(CVscore)
```
jaidevd, line 131: If `cv` is `True`, or unspecified, here it will pass to `cross_val_score` as `cv=True` - which is not a valid input.

PR #417, opened 2021-06-21
Add code for sentiment analysis

MSanKeys963:
Removed `class NLPHandler()` and added sentiment analysis functionality in `class MLHandler()`.

To setup a Gramex service for performing sentiment analysis, use the following configuration:

```yaml
url:
  sentiment-analysis:
    pattern: /$YAMLURL/
    handler: MLHandler
    kwargs:
      backend: transformers
      task: sentiment-analysis
      xsrf_cookies: false
```

## Getting predictions

`GET` sentiments of short pieces of text as follows:

```bash
curl -X GET --data-urlencode "text=This movie is so bad, it's good." http://localhost:9988/
```

The output will be:
```json
[
  {
    "label": "POSITIVE",
    "score": 0.9997316002845764
  }
]
```

Files containing text to be classified can also be `POST`ed to the endpoint, with `_action=predict`. Any file supported by `gramex.cache.open` will work. (Download a sample [here](https://drive.google.com/file/d/146MXfbBNpC8xI0Txpra3xs25WHpHk73k/view?usp=sharing).)

```bash
curl -X POST -F "file=@sentiment.csv" http://localhost:9988/?_action=predict
```

The output will be:

```json
[
  {
    "label": "POSITIVE",
    "score": 0.9997316002845764
  },
  {
    "label": "NEGATIVE",
    "score": 0.9974692463874817
  },
  // etc.
]
```

## Measuring model performance

Files containing the `text` and `label` fields can be `POSTED` to the endpoint
with `_action=score` to get the [ROC AUC score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html) of the model against the dataset. (Download a sample dataset [here](https://drive.google.com/file/d/146MXfbBNpC8xI0Txpra3xs25WHpHk73k/view?usp=sharing)).

```bash
curl -X POST -F "file=@sentiment.csv" http://localhost:9988/?_action=score
```

The output will be something like:

```json
{
  "roc_auc": 0.9929
}
```

## Training the model

The model can be trained on a dataset by setting `_action=train`, and `POST`ing the file.

```bash
curl -X POST -F "file=@sentiment_score.json" http://localhost:9988/?_action=train
```

The output will show the score of the trained model on the dataset:

```json
{
  "roc_auc": 0.8
}
```

Multiple training options for the transformer are supported, including the number of epochs, batch size and weight decay. These can all be specified in the `POST` request as follows:

```bash
# Train for three epochs instead of the default 1
curl -X POST -F "file=@sentiment.csv" http://localhost:9988/?_action=train&num_train_epochs=3
```

The output is the score of the trained model on the dataset after 3 epochs:

```json
{
  "roc_auc": 0.98
}
```

```bash
# Change the batch size to 32 instead of the default 16
curl -X POST -F "file=@sentiment.csv" \
	http://localhost:9988/?_action=train&per_device_train_batch_size=32&num_train_epochs=3
```

The output is the score of the trained model on the dataset after 3 epochs and a batch size of 32:

```json
{
  "roc_auc": 0.99
}
```

### DIFF
--- a/gramex/handlers/mlhandler.py
+++ b/gramex/handlers/mlhandler.py
@@ -524,7 +524,6 @@ class MLHandler(BaseMLHandler):
             if action == 'train':
                 kwargs = self._coerce_transformers_opts()
                 kwargs['model_path'] = self.model_path
-                print(self.model_path)
                 args = _train_transformer, self.model, data
             elif action == 'score':
                 args = _score_transformer, self.model, data
--- a/gramex/handlers/mlhandler.py
+++ b/gramex/handlers/mlhandler.py
@@ -27,6 +27,7 @@ try:
     from transformers import AutoModelForSequenceClassification, AutoTokenizer
     from transformers import Trainer, TrainingArguments
     from gramex.dl_utils import SentimentDataset
+    TRANSFORMERS_INSTALLED = True
 except ImportError:
     TRANSFORMERS_INSTALLED = False
 
@@ -248,6 +249,7 @@ class BaseMLHandler(FormHandler):
 
 
 class MLHandler(BaseMLHandler):
+    model_backend = ""
 
     @classmethod
     def setup(cls, data=None, model={}, backend='sklearn', config_dir='', **kwargs):
@@ -262,13 +264,11 @@ class MLHandler(BaseMLHandler):
         # elif backend == 'transformers':
         #     NLPHandler.fit(**kwargs)
         if backend != 'sklearn':
+            cls.model_backend = backend
             if not TRANSFORMERS_INSTALLED:
                 raise ImportError('pip install transformers')
             super(MLHandler, cls).setup(**kwargs)
             cls.load_transformer(task, model)
-            cls.get = NLPHandler.get
-            cls.post = NLPHandler.post
-            cls.delete = NLPHandler.delete
         else:
             super(MLHandler, cls).setup(data, model, config_dir, **kwargs)
             # Handle data if provided in the YAML config.
@@ -450,6 +450,10 @@ class MLHandler(BaseMLHandler):
             self.write(json.dumps(params, indent=2))
         elif '_cache' in self.args:
             self.write(self.load_data().to_json(orient='records'))
+        elif 'text' in self.args:
+            text = self.get_arguments('text')
+            result = yield gramex.service.threadpool.submit(self.model, text)
+            self.write(json.dumps(result, indent=2))
         else:
             self._check_model_path()
             if '_download' in self.args:
@@ -513,8 +517,24 @@ class MLHandler(BaseMLHandler):
         action = self.args.pop('_action', 'predict')
         if action not in ACTIONS:
             raise ValueError(f'Action {action} not supported.')
-        res = yield gramex.service.threadpool.submit(getattr(self, f"_{action}"))
-        self.write(json.dumps(res, indent=2, cls=CustomJSONEncoder))
+        if self.model_backend == "transformers":
+            data = self._parse_data(_cache=False)
+            move_to_cpu(self.model)
+            kwargs = {}
+            if action == 'train':
+                kwargs = self._coerce_transformers_opts()
+                kwargs['model_path'] = self.model_path
+                print(self.model_path)
+                args = _train_transformer, self.model, data
+            elif action == 'score':
+                args = _score_transformer, self.model, data
+            elif action == 'predict':
+                args = self.model, data['text'].tolist()
+            res = yield gramex.service.threadpool.submit(*args, **kwargs)
+            self.write(json.dumps(res, indent=2, cls=CustomJSONEncoder))
+        else:
+            res = yield gramex.service.threadpool.submit(getattr(self, f"_{action}"))
+            self.write(json.dumps(res, indent=2, cls=CustomJSONEncoder))      
         super(MLHandler, self).post(*path_args, **path_kwargs)
 
     def get_cached_arg(self, argname):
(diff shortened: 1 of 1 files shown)


sanand0:
Cool! @jaidevd could you please review? Do let me know when to merge

jaidevd:
@MSanKeys963 The target branch has to be `gramener/gramex`'s master branch, not the `jd-transformers` branch.

gramex/handlers/mlhandler.py
```diff
@@ -450,6 +450,10 @@ def get(self, *path_args, **path_kwargs):
             self.write(json.dumps(params, indent=2))
         elif '_cache' in self.args:
             self.write(self.load_data().to_json(orient='records'))
+        elif 'text' in self.args:
```
jaidevd, line 453: This condition doesn't check `backend`. Which means, that even if I'm running an sklearn model which has one column named `text`, and I request predictions from this model - this piece of code will run.

gramex/handlers/mlhandler.py
```diff
@@ -262,13 +264,11 @@ def setup(cls, data=None, model={}, backend='sklearn', config_dir='', **kwargs):
         # elif backend == 'transformers':
         #     NLPHandler.fit(**kwargs)
         if backend != 'sklearn':
+            cls.model_backend = backend
```
jaidevd, line 267: Do we need this extra variable? `backend` can imply be assigned to `cls.backend`.

jaidevd (changes requested):

jaidevd:
@MSanKeys963 other than these two changes, LGTM

jaidevd:
@MSanKeys963 this still showing merge conflicts. Please take a look.

gramex/handlers/formhandler.py
```diff
@@ -53,6 +53,7 @@ def setup(cls, **kwargs):
         cls.headers = conf_kwargs.pop('headers', {})
         # Top level formats: key is special. Don't treat it as data
         cls.formats = conf_kwargs.pop('formats', {})
+        cls.task = conf_kwargs.pop('task', {})
```
jaidevd, line 56: @MSanKeys963 We should not have to make any changes to FormHandler.

gramex/install.py
```diff
@@ -197,32 +197,42 @@ def _ensure_remove(function, path, exc_info):
...
+            # auto-resuming rmtree: https://bugs.python.org/issue8523
+            _try_remove(target, retries, delay, func, **kwargs)
+            return True
+        else:
+            app_log.warning('Not removing directory %s (outside $GRAMEXDATA)', target)
+            return False
     else:
-        app_log.warning('Not removing directory %s (outside $GRAMEXDATA)', target)
-        return False
+        _try_remove(target, retries, delay, func, **kwargs)
```
jaidevd, line 236: This file should not have changed. Please replace this file with the version available on master.

jaidevd (changes requested):

gramex/handlers/formhandler.py
MSanKeys963: Can we keep this line?
It takes care of `task: sentiment-analysis` from `gramex.yaml`.

jaidevd: Why do we need `cls.task` in FormHandler? It doesn't mean anything to the FormHandler class. This attribute is just left here, and the class doesn't do anything with it. Also, if it's just a matter of popping the task key from the `conf_kwargs` dict, that can easily be done in `MLHandler` itself, before calling `super(MLHandler, ...)`

gramex/handlers/mlhandler.py
```diff
@@ -369,9 +452,29 @@ def post(self, *path_args, **path_kwargs):
...
+            elif action == 'predict':
+                args = self.model, data['text'].tolist()
+            res = yield gramex.service.threadpool.submit(*args, **kwargs)
+            self.write(json.dumps(res, indent=2, cls=CustomJSONEncoder))
+        else:
+            res = yield gramex.service.threadpool.submit(getattr(self, f"_{action}"))
+            self.write(json.dumps(res, indent=2, cls=CustomJSONEncoder))      
         super(MLHandler, self).post(*path_args, **path_kwargs)
+            
+    def get_cached_arg(self, argname):
```
jaidevd, line 474: It turns out that we don't need this function either. It's not being used anywhere. Please remove this as well.

gramex/handlers/formhandler.py
MSanKeys963: Got it. Thank you!

MSanKeys963:
@jaidevd I've fixed all the issues mentioned above. Please let me know if there's anything else.

jaidevd:
Thanks, @MSanKeys963 

@sanand0 This is ready for merge.

gramex/dl_utils.py
```diff
@@ -0,0 +1,16 @@
+from torch.utils.data import Dataset
```
sanand0, line 1: Let's rename this file to `dl.py`

gramex/handlers/mlhandler.py
```diff
@@ -16,11 +16,20 @@
...
 from sklearn.pipeline import Pipeline
-from sklearn.preprocessing import OneHotEncoder, StandardScaler
+from sklearn.metrics import roc_auc_score
+from sklearn.preprocessing import OneHotEncoder, StandardScaler, LabelEncoder
 from slugify import slugify
 from tornado.gen import coroutine
 from tornado.web import HTTPError
 from sklearn.metrics import get_scorer
+try:
+    from transformers import pipeline
```
sanand0, line 26: Let's have a consistent policy for imports. Treat torch and huggingface the same way.

MSanKeys963: @sanand0  Fixed it.

MSanKeys963:
@sanand0 I've fixed all the issues. Please check.

sanand0:
@MSanKeys963

- [ ] Can you get this to work, please? [sentiment.zip](https://github.com/gramener/gramex/files/6905106/sentiment.zip)
- [ ] Gramex should still run if PyTorch & Huggingface are not installed

For example, this is how we optionally import ElasticSearch:

```python
def gramexlog(conf):
    try:
        from elasticsearch import Elasticsearch, helpers
    except ImportError:
        app_log.error('gramexlog: elasticsearch missing. pip install elasticsearch')
        return
```

PR #423, opened 2021-07-15
GRAMEX-40 ⁃ ENH: Support sklearn decompositions in MLHandler

jaidevd:
Use the following configuration:

```yaml
url:
  ml-home:
    pattern: /$YAMLURL/
    handler: MLHandler
    kwargs:
      data:
        url: $YAMLPATH/iris.csv
      model:
        class: PCA  # Or TruncatedSVD, or any other class that does dimensionality reduction
        exclude: [species]
        params:
                n_components: 2
      xsrf_cookies: false
```

┆Issue is synchronized with this [Jira Bug](https://gramenertech.atlassian.net/browse/GRAMEX-40)


gramex/handlers/mlhandler.py
```diff
@@ -272,12 +285,16 @@ def _predict(self, data=None, score_col=''):
             if metric:
                 scorer = get_scorer(metric)
                 return scorer(self.model, data, target)
-            return self.model.score(data, target)
+            score = self.model.score(data, target)
+            return score
```
jaidevd, line 289: What?

gramex/handlers/mlhandler.py
```diff
@@ -31,6 +32,14 @@
...
     'sklearn.neural_network',
     'sklearn.naive_bayes',
+    'sklearn.decomposition'
+
+]
+SKLEARN_DECOMPOSE_ATTRS = [
+    "components_",
+    "explained_variance_",
+    "explained_variance_ratio_",
+    "singular_values_"
```
sanand0, line 42: Could you please check what other attributes other models provide. They may provide more. Or less. Could we expose all model attributes rather than pre-selecting?

gramex/handlers/mlhandler.py
```diff
@@ -348,11 +364,16 @@ def _train(self, data=None):
...
-        _fit(self.model, train, target, self.model_path)
-        return {'score': self.model.score(train, target)}
+        if not isinstance(self.model[-1], TransformerMixin):
+            target = data[target_col]
+            train = data[[c for c in data if c != target_col]]
+            _fit(self.model, train, target, self.model_path)
+            result = {'score': self.model.score(train, target)}
+        else:
+            _fit(self.model, data, path=self.model_path)
+            result = {k: getattr(self.model[-1], k) for k in SKLEARN_DECOMPOSE_ATTRS}
```
sanand0, line 375: Should we just return whatever attributes `.model` has? This should work with *any* TransformerMixin.

gramex/handlers/mlhandler.py
jaidevd: Yes, there are many classes in `sklearn.decomposition` that have attributes other than these. By convention they all have names that end in `_`. We'll just return any attribute a model has that ends in an underscore.

jaidevd: @sanand0  Done, please take a look at the last commit.

jaidevd:
@sanand0 This PR shows how to access fitted attributes of sklearn models.

jaidevd:
@sanand0 `GET /?_params` now returns fitted attributes whenever they are available. Please take a look.

PR #456, opened 2021-09-28
ENH: Update multiple rows in files and DBs with data.update

### LINKED ISSUE #312: Cannot update multiple rows with gramex.data.update
**Is something not working as expected?**
`gramex.data.update` does not parse `args` correctly. 

**Steps to reproduce.**
Suppose there's a `file.csv` containing:
```
id, val
1, x
2, y
3, z
```
then the following:
```python
gramex.data.update('file.csv', id="id", args={'id': [2, 3], 'val': ['a', 'b']})
```
converts it into:
```
id, val
1, x
2, a
3, a
```

**Expected behavior**
It should update rows 2 and 3 to `a` and `b` _respectively_.

**Your environment:**
 - Gramex version: master branch as of this bug report.

jaidevd:
Closes #312

### DIFF
--- a/testlib/test_data.py
+++ b/testlib/test_data.py
@@ -597,6 +597,49 @@ class TestEdit(unittest.TestCase):
         gramex.data.update(data, args=args, id=['देश', 'city', 'product'])
         ase(types_original, data.dtypes)
 
+    def test_update_multiple_file(self):
+        # Test on a file
+
+        update_file = os.path.join(folder, 'actors.update.csv')
+        shutil.copy(os.path.join(folder, '..', 'tests/actors.csv'), update_file)
+        self.tmpfiles.append(update_file)
+
+        names = ['Humphrey Bogart', 'James Stewart', 'Audrey Hepburn']
+        categories = ['Stars', 'Thespians', 'Heartthrobs']
+        ratings = [1, 0.99, 1.11]
+        gramex.data.update(
+            update_file,
+            args={
+                'name': names,
+                'category': categories,
+                'rating': ratings
+            }, id=['name']
+        )
+        df = gramex.data.filter(update_file, args={'name': names})
+        self.assertEqual(df['category'].tolist(), categories)
+        self.assertEqual(df['rating'].tolist(), ratings)
+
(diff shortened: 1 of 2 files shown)


testlib/test_data.py
```diff
@@ -597,6 +597,49 @@ def test_update(self):
...
+
+    def test_update_multiple_db(self):
+        actors = gramex.cache.open(os.path.join(folder, '../tests/actors.csv'))
+        temp_db = f'sqlite:///{folder}/actors.db'
+        self.tmpfiles.append(os.path.join(folder, 'actors.db'))
+        actors.to_sql('actors', sa.create_engine(temp_db), index=False)
+
+        names = ['Humphrey Bogart', 'James Stewart', 'Audrey Hepburn']
+        categories = ['Stars', 'Thespians', 'Heartthrobs']
+        ratings = [1, 0.99, 1.11]
```
sanand0, line 630: Please add a test case where 

- `ratings` has only 2 values [1, 0.99] and the third is missing. Audrey's rating should not get updated, but her categories should be.
- `names` has only 2 values -- in which case, the 3rd category and rating are ignored

testlib/test_data.py
```diff
@@ -597,6 +597,49 @@ def test_update(self):
...
+            args={
+                'name': names,
+                'category': categories,
+                'rating': ratings
+            }, id=['name'], table='actors'
+        )
+        df = gramex.data.filter(temp_db, args={'name': names}, table='actors')
+        self.assertEqual(df['category'].tolist(), categories)
+        self.assertEqual(df['rating'].tolist(), ratings)
+
```
sanand0, line 642: Please add test cases for MySQL, PostgreSQL.

testlib/test_data.py
jaidevd: @sanand0 This works for files, but for sqlalchemy, to deal with uneven args, we might have to do column-wise update queries. Is this acceptable?

sanand0: @jaidevd -- yes, column-wise update queries are fine. I don't anticipate this to be used often and we can optimize later

PR #460, opened 2021-10-07
ENH: Modify user attributes with rules

jaidevd:
Use the following spec:

```yaml
url:
    auth/db:
    pattern: /$YAMLURL/db
    handler: DBAuth
    kwargs:
      url: $YAMLPATH/users.csv
      user:
        column: user
      password:
        column: password
      redirect:
        query: next
        header: Referer
        url: .
      rules:
        url: $YAMLPATH/rules.csv
```

And the following two files:

1. [users.csv](https://github.com/gramener/gramex/files/7299389/users.csv)

2. [`rules.csv`](https://github.com/gramener/gramex/files/7299388/rules.csv)

When logging in, each matching rule will be applied to the current user.

### DIFF
--- a/gramex/handlers/authhandler.py
+++ b/gramex/handlers/authhandler.py
@@ -2,6 +2,7 @@ import io
 import os
 import csv
 import json
+import re
 import time
 import uuid
 import logging
@@ -32,10 +33,12 @@ class AuthHandler(BaseHandler):
 
     @classmethod
     def setup(cls, prepare=None, action=None, delay=None, session_expiry=None,
-              session_inactive=None, user_key='user', lookup=None, recaptcha=None, **kwargs):
+              session_inactive=None, user_key='user', lookup=None, recaptcha=None,
+              rules=None, **kwargs):
         # Set up default redirection based on ?next=...
         if 'redirect' not in kwargs:
             kwargs['redirect'] = AttrDict([('query', 'next'), ('header', 'Referer')])
+        cls.special_keys += ['rules']
         super(AuthHandler, cls).setup(**kwargs)
 
         # Set up logging for login/logout events
@@ -69,6 +72,10 @@ class AuthHandler(BaseHandler):
                 cls.lookup_id = cls.lookup.pop('id', 'user')
             else:
                 app_log.error('%s: lookup must be a dict, not %s', cls.name, cls.lookup)
+
+        if rules:
+            rules = gramex.data.filter(**rules)
+        cls.rules = rules
 
         # Set up prepare
         cls.auth_methods = {}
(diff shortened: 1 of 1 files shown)


gramex/handlers/authhandler.py
```diff
@@ -70,6 +73,10 @@ def setup(cls, prepare=None, action=None, delay=None, session_expiry=None,
             else:
                 app_log.error('%s: lookup must be a dict, not %s', cls.name, cls.lookup)
 
+        if rules:
+            rules = gramex.data.filter(**rules)
+        cls.rules = rules
```
jaidevd, line 78: If rules is None, create an empty dataframe.

tests/test_auth.py
```diff
@@ -483,6 +483,62 @@ def test_lookup(self):
         eq_(session['user']['gender'], 'female')
 
 
+class TestRules(AuthBase):
+
+    @classmethod
+    def setUpClass(cls):
```
sanand0, line 489: Another approach is to just set url = server.base_url + '/auth/rulesdb' as a class variable and not override setupClass at all. Same for TestRulesFile

tests/gramex.yaml
```diff
@@ -1993,6 +1993,70 @@ url:
...
+          password: beta
+          gender: male
+          email: john.doe@gramener.com
+          team: "ग्रामेक्स"
+        gamma:
+          password: gamma
+          gender: male
+          empty: ''
+      rules:
+        url: $YAMLPATH/rules.xlsx
```
sanand0: Please make this a rules.csv -- that way, at least one file is inspectable

tests/gramex.yaml
```diff
@@ -1993,6 +1993,70 @@ url:
...
+        alpha:
+          password: alpha
+          gender: female
+          email: jane.doe@gmail.com
+        beta:
+          password: beta
+          gender: male
+          email: john.doe@gramener.com
+          team: "ग्रामेक्स"
+        gamma:
```
sanand0, line 2032: Could this username contain unicode instead of gamma?

tests/test_auth.py
```diff
@@ -483,6 +483,62 @@ def test_lookup(self):
...
+        eq_(session['user']['gender'], 'female')
+        eq_(session['user']['team'], 'Gramex')
+
+        # When selector is a nonexistent attribute, nothing changes
+        self.login_ok('gamma', 'gamma', check_next='/')
+        session = self.session.get(server.base_url + '/auth/session').json()
+        eq_(session['user']['gender'], 'female')
+
+        # Check if the empty string match for Gamma works
+        eq_(session['user']['email'], 'gamma@null.com')
```
sanand0, line 509: Add test case where we can add field="expiry" with a value= of date type. Also a field="cookie_expires" with value= of float type

sanand0: Please also test that no user has an email none@none.com nor gamma@null.com.

Also test that team=Gramex for all

PR #462, opened 2021-10-08
GRAMEX-103 ⁃ FIX: FormHandler query not working with MongoDB

radheyakale:
┆Issue is synchronized with this [Jira Bug](https://gramenertech.atlassian.net/browse/GRAMEX-103)


gramex/data.py
```diff
@@ -13,6 +13,8 @@
 from tornado.escape import json_encode
 from gramex.config import merge, app_log
 from orderedattrdict import AttrDict
+from bson.objectid import ObjectId
```
sanand0, line 16: Don't import bson at the top level. Gramex does not require MongoDB. If it's there, we use it. So import it inside the MongoDB filter functions where required.

gramex/data.py
```diff
@@ -212,6 +214,8 @@ def filter(url, args={}, meta={}, engine=None, ext=None, columns=None,
     })
     controls = _pop_controls(args)
     transform = _transform_fn(transform, transform_kwargs)
+    if type(query) == str and 'url' in args.keys() and args['url'].startswith('mongodb'):
```
sanand0, line 217: We don't want anything mongodb (or any DB) specific in filter(). That should be part of the plugins only.

gramex/data.py
```diff
@@ -1460,9 +1464,24 @@ def _mongodb_json(obj):
...
+        if key == '_id':
+            args[key] = list(map(ObjectId, value))
+            continue
+
+        for val in value:
+            if not val.isnumeric():
+                break
+        else:
+            args[key] = list(map(int, value))
+
```
sanand0, line 1478: You can get the first 100 rows from `meta_cols = pd.DataFrame(table.find().limit(100))`. Use the DataFrame dtype to figure out the type and convert based on that.

There's a line that goes `convert = int if (meta_cols[col].dtype == pd.np.int64) else meta_cols[col].dtype.type`

If you plug into that line and extend it for these scenarios, we should be OK

gramex/data.py
```diff
@@ -1460,9 +1464,24 @@ def _mongodb_json(obj):
 def _filter_mongodb(url, controls, args, database=None, collection=None, query=None, **kwargs):
     '''TODO: Document function and usage'''
     table = _mongodb_collection(url, database, collection, **kwargs)
+
+    for key, value in args.items():
+        if key == '_id':
```
sanand0, line 1469: We don't want to depend on the column name. Depend on the actual type.

gramex/data.py
sanand0: BTW, we don't want to eval it. Just take the object as-is and make it work.

gramex/data.py
```diff
@@ -1381,9 +1382,16 @@ def _type_conversion(param_list, operations=None):
 }
 
 
-def _filter_mongodb_col(col, op, vals, meta_cols):
+def _filter_mongodb_col(col, op, vals, meta_cols, object_key=None):
+    import bson
+
     if op in ['', '!']:
-        return {col: {_mongodb_op_map[op]: vals}}
+        convert = int if (meta_cols[col].dtype == pd.np.int64) else \
```
sanand0, line 1389: This conversion needs to apply even if the value is `>` or `<` etc. Convert the type for all values based on the type.

gramex/data.py
```diff
@@ -1381,9 +1382,16 @@ def _type_conversion(param_list, operations=None):
...
+            float if (meta_cols[col].dtype == pd.np.float64) else \
```
sanand0, line 1390: Are we handling all relevant types:
- What about datetimes?
- Will pymongo ever return a type other than pd.np.float64 for float?
- Will pymongo ever return a type other than pd.np.int64 for int?
- What about booleans?

gramex/data.py
```diff
@@ -1471,6 +1492,7 @@ def _filter_mongodb(url, controls, args, database=None, collection=None, query=N
         for col, val in data.iloc[0].iteritems():
             if type(val) in {bson.objectid.ObjectId}:
                 data[col] = data[col].map(str)
+
```
sanand0, line 1495: Let's add test cases in test_lib/test_data.py.

But manually also test for

- Filter by ID
- Filter a table that has multiple ID columns using two IDs
- Filter a table with a boolean, date, int, float and str column with EACH of these
- Delete by ID, date, boolean, int, str
- Update one or more fields filtered by ID, date, boolean, int, str
- Test Gramex Admin with a MongoDB backend: add user/update password/delete users

gramex/data.py
```diff
@@ -1394,15 +1402,24 @@ def _filter_mongodb_col(col, op, vals, meta_cols):
...
     # Convert a query like x>=3&x>=4&x>=5 into
     # {"$or": [{x: {$gt: 3}}, {x: {$gt: 4}}, {x: $gt: 5}]}
     # TODO: ?_id= is not working
+    import bson
+
+    object_key = None
+    if results:
+        for k, v in results[0].items():
+            if type(v) == bson.objectid.ObjectId:
+                object_key = k
```
sanand0, line 1415: What if there are multiple object keys? Make this a set, not a single column name. Name it object_keys. Add a comment explaining that Pandas returns strings for both.

Another possibility. Don't use Pandas to read the first few rows. Instead, use pymongo & detect the type directly. This may be better.

gramex/data.py
```diff
@@ -1460,9 +1477,13 @@ def _mongodb_json(obj):
...
-    if query is None:
-        meta_cols = pd.DataFrame(list(table.find().limit(100)))
-        query = _mongodb_query(args, meta_cols)
+
+    results = table.find().limit(100)
+    meta_cols = pd.DataFrame(list(copy.deepcopy(results)))
+    if query is not None:
+        query = dict(query)
+    else:
+        query = _mongodb_query(args, meta_cols, results=results)
```
sanand0, line 1486: Again, another possibility. Don't use Pandas to read the first few rows. Instead, use pymongo & detect the type directly. This may be better.

sanand0:
@nikhilkabbin  -- could you please review this to see if it will meet your needs and if any scenarios won't be handled? Please discuss with @radheyakale

gramex/data.py
```diff
@@ -1481,23 +1524,41 @@ def _filter_mongodb(url, controls, args, database=None, collection=None, query=N
...
+    values = {key: val[0] for key, val in dict(args).items() if key not in id}
+
+    for key, val in values.items():
+        convert = type(results[0][key])
+
+        if convert == datetime:
+            convert = parse
+
+        if convert == bool:
+            values[key] = 0 if val in ['0', 'false', 'False'] else 1
```
sanand0, line 1558: Don't repeat yourself. Have a single convert function to do this and use it in `_filter_mongodb_col`

gramex/data.py
```diff
@@ -1391,28 +1392,66 @@ def _type_conversion(param_list, operations=None):
...
         return {col: {"$regex": '|'.join(vals), "$options": 'i'}}
     elif col and op in _mongodb_op_map.keys():
         # TODO: Improve the numpy to Python type
-        convert = int if (meta_cols[col].dtype == pd.np.int64) else meta_cols[col].dtype.type
+        convert = type(results[0][col])
         return {col: {_mongodb_op_map[op]: convert(val)} for val in vals}
 
 
-def _mongodb_query(args, meta_cols):
+def _mongodb_query(args, meta_cols, results=None, **kwargs):
```
sanand0, line 1438: Don't pass results here. Instead, take table. Set `results = table.find().limit(1)` -- we're not using more rows. Handle scenario where there are no rows, i.e. no data

gramex/data.py
```diff
@@ -1470,9 +1509,13 @@ def _mongodb_json(obj):
 def _filter_mongodb(url, controls, args, database=None, collection=None, query=None, **kwargs):
     '''TODO: Document function and usage'''
     table = _mongodb_collection(url, database, collection, **kwargs)
-    if query is None:
-        meta_cols = pd.DataFrame(list(table.find().limit(100)))
-        query = _mongodb_query(args, meta_cols)
+
+    results = table.find().limit(100)
+    meta_cols = pd.DataFrame(list(copy.deepcopy(results)))
```
sanand0, line 1514: Check this out. `meta_cols` is being used only for

- `_controls_default` -- which just requires a list of column names
- `_filter_col`  -- which just requires a list of column names

So we don't need the DataFrame

gramex/data.py
```diff
@@ -1467,12 +1505,26 @@ def _mongodb_json(obj):
...
 
 
+def _convert_datatype(results, key):
+    from dateutil.parser import parse
+
+    convert = type(results[0][key])
+
+    if convert == datetime:
+        convert = parse
+
```
sanand0, line 1515:
```python
    elif convert == bool:
        convert = lambda val: 0 if val in ['0', 'false', 'False'] else 1
```

gramex/data.py
```diff
@@ -1481,23 +1533,39 @@ def _filter_mongodb(url, controls, args, database=None, collection=None, query=N
...
+        return 0
+
+    values = {key: val[0] for key, val in dict(args).items() if key not in id}
+
+    for key, val in values.items():
+        convert = _convert_datatype(results, key)
+
+        if convert == bool:
+            values[key] = 0 if val in ['0', 'false', 'False'] else 1
+
```
sanand0, line 1566: Reduce the number of empty newlines. Group into logical chunks, but slighly larger chunks than you have now :-)

gramex/data.py
```diff
@@ -1391,28 +1392,65 @@ def _type_conversion(param_list, operations=None):
...
     for key, vals in args.items():
-        col, agg, op = _filter_col(key, meta_cols)
+        if len(id) and key not in id:
+            continue
+
+        col_names = [k for k in results[0].keys()]
+        col, agg, op = _filter_col(key, col_names)
         if col:
-            conditions.append(_filter_mongodb_col(col, op, vals, meta_cols))
+            if results:
```
sanand0, line 1451: Can be written as `if col and results`

gramex/data.py
```diff
@@ -1391,28 +1392,65 @@ def _type_conversion(param_list, operations=None):
...
+def _filter_mongodb_col(col, op, vals, results, object_keys=[]):
+    from dateutil.parser import parse
+
     if op in ['', '!']:
-        return {col: {_mongodb_op_map[op]: vals}}
+        convert = _convert_datatype(results, col)
+
+        if convert == bool:
+            for index, val in enumerate(vals):
+                vals[index] = 0 if val in ['0', 'false', 'False'] else 1
```
sanand0, line 1403: This will vanish once you add the bool conversion logic into _convert_datatype

radheyakale:
Count in MongoDB is taken by data.count(). I would suggest the following diff:

`diff --git a/gramex/data.py b/gramex/data.py
index 56e80375..0e00bd4a 100644
--- a/gramex/data.py
+++ b/gramex/data.py
@@ -604,7 +604,10 @@ def _convert_datatype(data, col):
     if isinstance(data, pd.DataFrame):
         conv = data[col].dtype.type
     else:
-        conv = type(data[0][col]) if len(data) else lambda v: v
+        try:
+            conv = type(data[0][col]) if len(data) else lambda v: v
+        except TypeError:
+            conv = type(data[0][col]) if data.count() else lambda v: v
 
     # Convert based on Pandas datatype. But for boolean, convert from string as below
     if conv in {pd.np.bool_, bool}:
`

sanand0:
@radheyakale -- I've fixed the issue with count in MongoDB taken by `data.count()`. I've made several other refactorings as well. Please review. If this works for you, I'll merge.

radheyakale:
@sanand0 It can be merged now

PR #467, opened 2021-10-12
GRAMEX-108 ⁃ ENH: Show TopCause results in MLHandler

jaidevd:
This PR adds topcause results as a FormHandler table in the MLHandler template.
Screenshot from 2021-10-12 13-02-24

┆Issue is synchronized with this [Jira Bug](https://gramenertech.atlassian.net/browse/GRAMEX-108)


sanand0:
Note: @sanand0 -- convert TopCause into a TransformMixin. Do a reset_index() at the end because we ignore the index.

Issue #468, opened 2021-10-12
GRAMEX-109 ⁃ ComicHandler does not have a default character


radheyakale:
**Steps to reproduce.**

1. In absence of arguments in URL, "This XML file does not appear to have any style information associated with it. The document tree is shown below." message is visible.

**Expected behavior**

1. The page should have a default character

┆Issue is synchronized with this [Jira Bug](https://gramenertech.atlassian.net/browse/GRAMEX-109)

sanand0:
Closed via bf40fda

PR #469, opened 2021-10-12
GRAMEX-110 ⁃ ComicHandler does not have a default character #468

radheyakale:
* Added default character in comichandler

┆Issue is synchronized with this [Jira Bug](https://gramenertech.atlassian.net/browse/GRAMEX-110)

### DIFF
--- a/gramex/handlers/comichandler.py
+++ b/gramex/handlers/comichandler.py
@@ -8,7 +8,7 @@ from gramex.config import variables
 from gramex.http import BAD_REQUEST, INTERNAL_SERVER_ERROR
 
 _info = {}
-default_character = {"name":"aryan","emotion":"angry","pose":"handsinpocket"}
+
 
 class ComicHandler(BaseHandler):
     @coroutine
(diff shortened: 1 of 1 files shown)


gramex/handlers/comichandler.py
```diff
@@ -8,7 +8,7 @@
 from gramex.http import BAD_REQUEST, INTERNAL_SERVER_ERROR
 
 _info = {}
-
+default_character = {"name":"aryan","emotion":"angry","pose":"handsinpocket"}
```
sanand0, line 11: Please run flake8 on this. Spaces after colon are required.

gramex/handlers/comichandler.py
```diff
@@ -17,7 +17,11 @@ def get(self):
             _info['node'] = Node(port=9967, cwd=os.path.join(variables['GRAMEXAPPS'], 'ui'))
         node = _info['node']
         # Take the last argument, i.e. ?name=dee&name=ava => {name: ava}
-        args = {key: vals[-1] for key, vals in self.args.items()}
+
+        if len(self.args.keys()) != 0:
```
sanand0, line 21: This can be simplified to `if len(self.args.keys())` -- no need for != 0

gramex/handlers/comichandler.py
```diff
@@ -17,7 +17,11 @@ def get(self):
...
+            args = {key: vals[-1] for key, vals in self.args.items()}
+        else:
+            args = default_character
```
sanand0, line 24: The default_character can be inlined here to save a line

sanand0 (commented):
Looks good, @radheyakale -- just a few minor tweaks.

Another issue is -- what if they specify ?name=aryan but don't specify the emotion=? Then nothing will be rendered.

We currently use a [slightly sophisticated logic for this](https://github.com/gramener/comicgen/blob/f60c9a8883e015ea33b777026420ec6b2b211270/script.js#L44-L55). I don't believe we need to replicate this.

So, for now, no need to action that.

PR #471, opened 2021-10-20
GRAMEX-116 ⁃ Don't watch files in init minimal

jaidevd:
┆Issue is synchronized with this [Jira Bug](https://gramenertech.atlassian.net/browse/GRAMEX-116)

### DIFF
--- a/gramex/apps/init/minimal/gramex.template.yaml
+++ b/gramex/apps/init/minimal/gramex.template.yaml
@@ -1,3 +1,6 @@
+app:
+  watch: false
+
 url:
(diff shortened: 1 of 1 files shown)


jaidevd:
By default the watches the subtrees of the following locations:
1. The `$YAMLPATH` corresponding to the current app
2. The `$YAMLPATH` corresponding to Gramex's default `gramex.yaml`
3. the `$YAMLPATH`s of _all imported apps!_

Of these, only no. 2 should be watched by default. Source code in other locations is not likely to change.

PR #478, opened 2021-11-16
GRAMEX-127 ⁃ REQ: Restrict xlrd to version <2

jaidevd:
Later versions drop support for all file types except .xls
Read more: https://github.com/python-excel/xlrd/blob/master/README.rst

┆Issue is synchronized with this [Jira Bug](https://gramenertech.atlassian.net/browse/GRAMEX-127)

### DIFF
--- a/gramex/release.json
+++ b/gramex/release.json
@@ -79,7 +79,7 @@
     "tornado==5.1.1":             "REQ: Web server",
     "watchdog>=0.8":              "REQ: Monitor file changes",
     "tzlocal":                    "TODO: Why is this required?",
-    "xlrd":                       "REQ: gramex.data.download()"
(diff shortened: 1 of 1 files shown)


sanand0:
How about we change the default gramex.cache.open engine to openpyxl instead?

jaidevd:
Is openpyxl already a dependency? If so, we can change the default engine.

Also, is xlrd being used for anything other than reading xlsx? If so, we still need this restriction.

jaidevd:
1. No, openpyxl is not a requirement. However, it is a default we enforce in gramex.cache.open (L151)
2. No, xlrd is not being used explicitly anywhere.
This PR is moot.

PR #510, opened 2022-02-28
GRAMEX-163 ⁃ ENH: Add login templates to `gramex init minimal`

jaidevd:
The minimal app is a little too minimal.
To use AuthHandlers on the IDE, users have to create login templates themselves.
This PR adds login templates for Simple/DBAuth and Email auth to the
miniaml app. Users now just have to select the login template.

┆Issue is synchronized with this [Jira Bug](https://gramenertech.atlassian.net/browse/GRAMEX-163)

### DIFF
--- a/gramex/apps/init/minimal/emailauth.template.html
+++ b/gramex/apps/init/minimal/emailauth.template.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Email Auth</title>
+  <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/css/bootstrap.min.css" rel="stylesheet">
+</head>
+<body>
+  <div class="container col-md-6 col-lg-4 col-xl-3">
+    <h1 class="text-center mt-5">Email Auth</h1>
+
+    {%! if error %}
+      <div class="alert alert-danger">
+        <strong>Error</strong>: {{! msg }}
+      </div>
+    {%! end %}
+
+    {%! if email is None and otp is None %}
+      <form method="POST">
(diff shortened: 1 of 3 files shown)


sanand0:
@jaidevd -- here's an alternate proposal. Let the minimal template be minimal. On the IDE, when someone selects Simple/DB/Email Auth, copy the relevant template.

That way, the app isn't polluted with files for those who don't require auth.

Your thoughts?

jaidevd:
@sanand0 Agreed, that way we don't need to do anything in Gramex.

But, the bottleneck is `emailauth.template.html` - it's a very specific file which, as far as I can tell, is not packaged with Gramex. 

So, I can just do the following - commit these three files to somewhere in the IDE app, and on each request of an authhandler, copy them to the relevant project.

Is this what you mean?

sanand0:
`gramexenterprise/handlers/auth.email.template.html` is the file we need for EmailAuth. It's part of Gramex Enterprise.

So, no need to commit the files on the IDE. They're already there. I suggest copying:

- `gramex/handlers/auth.template.html` for SimpleAuth and DBAuth
- `gramexenterprise/handlers/auth.email.template.html` for EmailAuth
- `gramexenterprise/handlers/auth.sms.template.html` for SMSAuth

In addition, you also need to always copy `gramex/handlers/auth.recaptcha.template.html` for these.

jaidevd:
Hmm, so in a sense a project / container on the IDE has to "request" these files for itself from it's own Gramex installation. Fine.

Also, I'll need to make sure that these templates are rendered correctly in the context of the current project. Is [this](https://github.com/gramener/gramex/blob/master/gramex/install.py#L643) what I'm looking for?

jaidevd:
@sanand0 I seem to have greatly misunderstood the situation. None of these handlers _need_ a template. It's the IDE that insists on having some field in the `template:` kwarg 😢

PR #519, opened 2022-04-05
GRAMEX-174 ⁃ FIX: MLHandler pipeline feature

jaidevd:
This commit fixes a bug in the sklearn pipeline created by MLHandler,
which was breaking the examples on the guide. We did not have a test for
this bug earlier, adding one.

┆Issue is synchronized with this [Jira Bug](https://gramenertech.atlassian.net/browse/GRAMEX-174)

### DIFF
--- a/gramex/handlers/mlhandler.py
+++ b/gramex/handlers/mlhandler.py
@@ -34,7 +34,7 @@ DEFAULT_TEMPLATE = op.join(op.dirname(__file__), '..', 'apps', 'mlhandler', 'tem
 def get_model(mclass: str, model_params: dict, **kwargs) -> ml.AbstractModel:
     if not mclass:
         return
-    if op.isfile(mclass):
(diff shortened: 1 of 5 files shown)


gramex/handlers/mlhandler.py
```diff
@@ -198,7 +198,7 @@ def _predict(self, data=None, score_col=''):
 
     def _check_model_path(self):
         try:
-            self.model = self.store.load_model()
+            self.model = get_model(self.store.model_path, {})
```
jaidevd, line 201: This is a problem: `get_model` can only raise `ImportError`s, not  `FileNotFoundError`s, making the following two lines unreachable. Breaks three tests.

jaidevd: This function should not exist, move this logic to the `ModelStore`.

PR #534, opened 2022-05-11
GRAMEX-189 ⁃ ENH: spaCy NLP functions

jaidevd:
This can be used with pipelines as follows:

```yaml
pipelines:
  spacy:
    stages:
      - name: data
        function: pandas.read_json("test.json")
      - name: parser
        function: gramex.nlp.load()
      - name: docs
        function: data["text"].map(parser)
      - name: ner
        function: docs.map(gramex.nlp.ner).tolist()
```

Then, triggering `gramex.pipelines['spacy']()` will:

1. read a file into a df
2. load the spacy parser
3. convert a text column from the data into a pandas series of spacy `Doc`s
4. Get named entities from all these docs.

### DIFF
--- a/gramex/nlp.py
+++ b/gramex/nlp.py
@@ -1,0 +1,41 @@
+import gramex
+import spacy
+from typing import Optional
+
+
+def load(model="en_core_web_sm") -> spacy.lang:
+    return gramex.cache.open(model, spacy.load)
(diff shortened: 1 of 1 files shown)


sanand0:
Closed in favor of 899ab832

PR #540, opened 2022-05-29
GRAMEX-197 ⁃ DOC: Switch API docs from Sphinx to mkdocs

sanand0:
Earlier, we used Sphinx to auto-document all Gramex classes and functions. This was hosted at https://learn.gramener.com/gramex/

But Sphinx is hard to set up, and uses reST - which is harder than Markdown.

This PR switches to [mkdocs](https://www.mkdocs.org/) and [mkdocstrings](https://mkdocstrings.github.io/) which are easier to set up and use Markdown.

┆Issue is synchronized with this [Jira Bug](https://gramenertech.atlassian.net/browse/GRAMEX-197)


sanand0:
Fixed in 0c24ddd7

PR #555, opened 2022-06-20
DEP: rmarkdown deprecated in FileHandler transforms

sanand0:
RMarkdown is rarely used with Gramex. It is very hard to support.
The underlying packages have several installation issues and
compatibility issues with NumPy. So we're phasing out support.

### DIFF
--- a/gramex/transforms/template.py
+++ b/gramex/transforms/template.py
@@ -101,6 +101,7 @@ def rmarkdown(content, handler=None, **kwargs):
 
     !!! Deprecated
         Support for R is being phased out in favor of Python ML libraries.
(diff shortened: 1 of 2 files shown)


gramex/transforms/template.py
```diff
@@ -115,7 +118,9 @@ def rmarkdown(content, handler=None, **kwargs):
     '''
     import gramex.ml
     import gramex.services
+    from gramex.config import app_log
 
+    app_log.warning('rmarkdown: transform deprecated')
```
jaidevd, line 123: Mention the version when this will no longer be available.

PR #556, opened 2022-06-20
REF: Use task script instead of Makefile

sanand0:
In release.json, move pip dependencies into lib. So now, there's only
lib (for pip) and conda for conda dependencies.

Format release.json with prettier.

Add eslint to package.json. Makes it cleaner to run lint tasks.


gramex/release.json
```diff
@@ -15,76 +14,69 @@
...
-      "cron-descriptor":          "OPT: admin/schedule to pretty-print cron",
-      "orderedattrdict>=1.6.0":   "REQ: OrderedDict with attr access for configs",
-      "tables":                   "REQ: HDF5 reading / writing",
-      "redis>=2.10.0":            "SRV: RedisStore"
-  },
-
   "conda#": "Packages that can only be installed via conda on Windows, not via pip",
   "conda": {
-    "line_profiler":              "OPT: conda: For gramex.debug"
+    "line_profiler": "OPT: conda: For gramex.debug"
```
jaidevd, line 80: Let's remove this too.

sanand0: Yes. Will do that in a subsequent PR.

PR #557, opened 2022-06-20
TST: If PhantomJS is not installed, skip capture tests

sanand0:
PhantomJS is deprecated. There's no point testing it any more.
If PhantomJS is not installed, the tests should not fail.

### DIFF
--- a/tests/test_capturehandler.py
+++ b/tests/test_capturehandler.py
@@ -83,6 +83,7 @@ class TestCaptureHandler(TestGramex):
 
     @classmethod
     def setupClass(cls):
+        # If PhantomJS is not installed, skip the entire test class
         if not paths['phantomjs']:
             raise SkipTest('phantomjs is not installed')
         cls.capture = get_capture('default', port=9402)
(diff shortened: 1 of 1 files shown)


tests/test_capturehandler.py
```diff
@@ -83,6 +83,8 @@ class TestCaptureHandler(TestGramex):
 
     @classmethod
     def setupClass(cls):
+        if not paths['phantomjs']:
+            raise SkipTest('phantomjs is not installed')
```
jaidevd, line 87: See if you can use `unittest.skipUnless` as a decorator for the class.

sanand0: I wasn't able to get `unittest.skipUnless` to work because

- `TestCaptureHandler` is the base class for `TestCaptureHandlerChrome`. If I skipUnless the first, it skips the second
- It still runs setupClass, though the [docs](https://docs.python.org/3/library/unittest.html) say that it won't. Might be a unittest - nosetests compatibility issue

I'll add a comment for now and proceed.

PR #559, opened 2022-06-20
Fix editorconfig, flake8 and bandit lint issues

sanand0:


sanand0:
Notes from review with @radheyakale 

- Bandit does security tests. We use `# nosec` to skip tests. Earlier, we had a blanket 
- The changes for editorconfig + flake8 are redundant. Instead, don't use editorconfig for Python files.

PR #560, opened 2022-06-21
ENH: Computer Vision pipelines

radheyakale:


radheyakale:
- [ ] TODO: Need a robust way to find if the MLHandler expects image data for training/testing
- [ ] TODO: Separate class is required for training and testing CV models
- [ ] TODO: Code has to be made generic to dynamically select Resnet50/VGG, etc. models
- [ ] TODO: Model training parameters viz. epochs, batch size have to be taken from the user with defaults set
- [ ] TODO: setup and get_model functions have to be made more scalable
- [ ] TODO: ModelStore __init__ has to be made scalable
- [ ] 
@jaidevd @sanand0 Please add/edit/delete any more TODOs that you feel may be required.

jaidevd:
@radheyakale

1. Start by creating a wrapper class for Keras apps which inherits from [this](https://github.com/gramener/gramex/blob/master/gramex/ml_api.py#L248). Refer to the `HFTransformer` class for an example.
2. Move as much code away from `MLHandler` as possible. It's okay if the example only supports ResNet50 for now.
3. Provide a sample `gramex.yaml` which shows the usage.

radheyakale:
> @radheyakale
> 
> 1. Start by creating a wrapper class for Keras apps which inherits from [this](https://github.com/gramener/gramex/blob/master/gramex/ml_api.py#L248). Refer to the `HFTransformer` class for an example.
> 2. Move as much code away from `MLHandler` as possible. It's okay if the example only supports ResNet50 for now.
> 3. Provide a sample `gramex.yaml` which shows the usage.

@jaidevd Please look at the recent commit https://github.com/gramener/gramex/pull/560/commits/d7791e629f4b535b7526cd2ca6b3e816a7ef912f. Is this the way we're looking to

gramex/handlers/mlhandler.py
```diff
@@ -95,12 +95,9 @@ def setup(cls, data=None, model={}, config_dir='', template=DEFAULT_TEMPLATE, **
         model_params = model.get('params', {})
         cls.store.dump('class', mclass)
         cls.store.dump('params', model_params)
-        if op.exists(cls.store.model_path):  # If the pkl exists, load it
-            if op.isdir(cls.store.model_path):
-                mclass, wrapper = ml.search_modelclass(mclass)
-                cls.model = locate(wrapper).from_disk(mclass, cls.store.model_path)
```
jaidevd, line 101: This condition is required for loading transformers. Why is this removed?

gramex/ml_api.py
```diff
@@ -46,6 +46,10 @@
         "statsmodels.tsa.statespace.sarimax",
     ],
     "gramex.ml_api.HFTransformer": ["gramex.transformers"],
+    "gramex.ml_api.KerasApplications": [
+        "tensorflow.keras.applications.vgg16",
+        "tensorflow.keras.applications.resnet50"
```
jaidevd, line 51: You can just leave this at `tensorflow.keras.applications`, because
```python
from tensorflow.keras.applications import *
```
covers everything

gramex/ml_api.py
```diff
@@ -203,8 +207,12 @@ class ModelStore(cache.JSONStore):
 
     def __init__(self, path, *args, **kwargs):
         _mkdir(path)
-        self.data_store = op.join(path, "data.h5")
-        self.model_path = op.join(path, op.basename(path) + ".pkl")
+        if op.exists(op.join(path, "data.h5")):
```
jaidevd, line 210: Why is this condition added? The `self.data_store` and `self.model_path` variables are declared here to let the model know the paths to files where training data and the model (respectively) would be stored. They don't need to exist at the beginning.

gramex/ml_api.py
```diff
@@ -397,19 +405,20 @@ def _predict(self, X, **kwargs):
...
             params = {"text_col": "text", "target_col": "label"}
         self.params = params
         self.kwargs = kwargs
 
-    @classmethod
-    def from_disk(cls, path, klass):
-        model = op.join(path, "model")
-        tokenizer = op.join(path, "tokenizer")
-        return cls(klass(model, tokenizer))
-
```
jaidevd, line 412: The two diffs above are unnecessary.

gramex/ml_api.py
```diff
@@ -426,3 +435,49 @@ def _predict(
     ):
         text = X["text"]
         return self.model.predict(text)
+
+
+class KerasApplications(AbstractModel):
```
jaidevd, line 440: Call this `KerasApplication` - singular.

gramex/ml_api.py
```diff
@@ -426,3 +435,49 @@ def _predict(
...
+        self.model = model
+        if params is None:
+            params = {}
+        self.params = params
+        self.kwargs = kwargs
+
+    @classmethod
+    def from_disk(cls, path, klass):
+        # Load model from disk
+        return cls
```
jaidevd, line 451: This should be something like:
```python
def from_disk(cls, path, klass):
    from tensorflow.keras.models import load_model
    return load_model(path)
```

gramex/ml_api.py
```diff
@@ -426,3 +435,49 @@ def _predict(
...
+        from tensorflow.keras.preprocessing import image
+
+        mclass, wrapper = search_modelclass(kwargs['mclass'])
+        module_imp = __import__(mclass.__module__, fromlist=SEARCH_MODULES[wrapper])
+        model = mclass(include_top=True,
+                       weights="imagenet",
+                       input_tensor=None,
+                       input_shape=None,
+                       pooling=None,
+                       classes=1000)
```
jaidevd, line 463: No... this will create a new model every time this method is called. `self.model` should exist by the time this function is reached. Check the `.predict` methods of the other classes.

gramex/ml_api.py
```diff
@@ -426,3 +435,49 @@ def _predict(
...
+    @classmethod
+    def from_disk(cls, path, klass):
+        # Load model from disk
+        return cls
+
+    def predict(self, data=None, **kwargs):
+        from tensorflow.keras.preprocessing import image
+
+        mclass, wrapper = search_modelclass(kwargs['mclass'])
+        module_imp = __import__(mclass.__module__, fromlist=SEARCH_MODULES[wrapper])
```
jaidevd, line 457: None of this is required. The `get_model` does all of this.

gramex/ml_api.py
```diff
@@ -426,3 +435,49 @@ def _predict(
...
+        model = mclass(include_top=True,
+                       weights="imagenet",
+                       input_tensor=None,
+                       input_shape=None,
+                       pooling=None,
+                       classes=1000)
+        x = image.img_to_array(data)
```
jaidevd, line 464: What is `data` here? This will work only if `data` is a [Pillow Image](https://pillow.readthedocs.io/en/stable/reference/Image.html#the-image-class). Are we enforcing this in MLHandler?

gramex/ml_api.py
jaidevd: Okay I guess what you are trying to do is as follows. If we have some Keras app, then the `preprocess_input` and `decode_predictions` should come from the same module.  This can be done as follows:
```python
from gramex.config import locate
preprocess_input = locate('preprocess_input', [model.__module__])
decode_predictions = locate('decode_predictions', [model.__module__])
```

(where model is the actual Keras model)

jaidevd: The point is to not import the whole module and only get the functions we need.

gramex/handlers/mlhandler.py
```diff
@@ -208,7 +207,8 @@ def _predict(self, data=None, score_col=''):
     def _check_model_path(self):
         try:
             klass, wrapper = ml.search_modelclass(self.store.load('class'))
-            self.model = locate(wrapper).from_disk(self.store.model_path, klass=klass)
+            if hasattr(self.store, 'model_path') and not op.isdir(self.store.model_path):
+                self.model = locate(wrapper).from_disk(self.store.model_path, klass=klass)
```
jaidevd, line 211: Hmm... ModelStore needs a better API - this condition looks too complex and unreadable. There should be some property of the store itself which tells you whatever you want to know.

Anyway... why is this required?

And you can write this simply as:
```python
if op.isdir(getattr(self.store, 'model_path', '')):
    do_stuff()
```

gramex/handlers/mlhandler.py
```diff
@@ -241,13 +241,26 @@ def get(self, *path_args, **path_kwargs):
...
+            elif len(self.request.files.keys()) and \
+                self.request.files['image'][0].content_type in \
+                    ['image/jpeg', 'image/jpg', 'image/png']:
+                if '_action' in self.args and self.args['_action'] == 'predict':
+                    import cv2
+                    import imutils
+                    data = imutils.resize(cv2.imdecode(np.fromstring(
+                        self.request.files['image'][0].body, np.uint8), cv2.IMREAD_UNCHANGED),
+                        width=224)
+                    data = cv2.resize(data, (224, 224))
```
jaidevd, line 260: 1. All of this logic should be in the wrapper class.
2. OpenCV should not be a dependency. For loading images from files / streams, use [`PIL.Image.open`](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.open)
3. For resizing, use `skimage.transform.resize` or `tf.image.resize` or `PIL.Image.resize`.
4. The target size after resizing should not be hardcoded to `[224, 224]` - there are other sizes in Keras apps too. For this, you can check the shape of the input tensor in the corresponding model with `model.input_shape`.

gramex/handlers/mlhandler.py
```diff
@@ -241,13 +241,26 @@ def get(self, *path_args, **path_kwargs):
...
+                self.set_header('Content-Type', 'application/octet-stream')
                 self.set_header('Content-Disposition',
                                 f'attachment; filename={op.basename(self.store.model_path)}')
                 with open(self.store.model_path, 'rb') as fout:
                     self.write(fout.read())
             elif '_model' in self.args:
                 self.write(json.dumps(self.model.get_params(), indent=2))
+            elif len(self.request.files.keys()) and \
+                self.request.files['image'][0].content_type in \
+                    ['image/jpeg', 'image/jpg', 'image/png']:
```
jaidevd, line 253: This check should not happen here. As far as possible, send the request payload blindly to the `_fit` or `_predict` methods. The wrapper class should take care of everything if it is written correctly.

Also note that there are two ways one can send images in a request.

1. Send files (multipart form data) - in which case you look at the mimetypes of the files received and open them accordingly. For this, take a look at the definition of `self._parse_multipart_form_data`
2. Send the raw bytestream of an image with a `Content-Type: image/whatever` header. In this case, write functions called `_parse_image_jpeg`, `_parse_image_png`, etc. The handler will automatically call them.

Basically the handler knows how to parse data given the content type of the request.

gramex/handlers/mlhandler.py
```diff
@@ -259,10 +272,15 @@ def get(self, *path_args, **path_kwargs):
...
                     data = []
                 if len(data) > 0:
-                    data = data.drop([self.store.load('target_col')], axis=1, errors='ignore')
-                    prediction = yield gramex.service.threadpool.submit(
-                        self._predict, data)
-                    self.write(json.dumps(prediction, indent=2, cls=CustomJSONEncoder))
+                    if 'training_data' in data.keys():
+                        training_results = yield gramex.service.threadpool.submit(
+                            self._train, data=data['training_data'].iloc[0])
+                        self.write(json.dumps(training_results, indent=2, cls=CustomJSONEncoder))
```
jaidevd, line 278: What does this do?

gramex/handlers/mlhandler.py
```diff
@@ -271,6 +289,61 @@ def get(self, *path_args, **path_kwargs):
...
+        keras_model.fit(
+            train_ds,
+            validation_data=val_ds,
+            epochs=epochs
+        )
+        with open(op.join(config_dir, 'class_names.json'), 'w') as fout:
+            json.dump(class_names, fout)
+        keras_model.save(config_dir)
+        return class_names
+
```
jaidevd, line 346: Remove this and move it to the wrapper class.

jaidevd (changes requested):

gramex/handlers/mlhandler.py
```diff
@@ -275,14 +291,18 @@ def _train(self, data=None):
...
+            return result
+        else:
+            data = self._parse_data(False) if data is None else data
+            data = self._filtercols(data)
+            data = self._filterrows(data)
+            self.model = get_model(
+                self.store.load('class'), self.store.load('params'),
+                data=data, target_col=target_col,
+                nums=self.store.load('nums'), cats=self.store.load('cats')
+            )
```
jaidevd, line 305: @jaidevd  - find a way to remove dataframe-specific code from MLHandler - it should deal only with train / predict semantics.

gramex/ml_api.py
jaidevd: Move this to `__init__`.

jaidevd: This should do only `self.model.predict`.

jaidevd (changes requested):

sanand0:
Closed as outdated

PR #577, opened 2022-07-18
REF: Model files are always stored at model.pkl

sanand0:


jaidevd:
@sanand0 LGTM, except we'll have to make a small change to the guide asking users to put their models at `somefolder/model.pkl` and use that as the `config_dir` kwarg. I'll send a PR for that tomorrow.

jaidevd (approved):

PR #578, opened 2022-07-18
ENH: Allow revoking OTP and API keys

sanand0:
handler.revoke_otp(otp) revokes the OTP.
handler.revoke_apikey(key) revokes the API key.

### DIFF
--- a/gramex/handlers/basehandler.py
+++ b/gramex/handlers/basehandler.py
@@ -702,9 +702,17 @@ class BaseMixin(object):
         self._session_store.dump(f'{prefix}:{otp}', {'user': user, '_t': time.time() + expire})
         return otp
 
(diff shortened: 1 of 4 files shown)


radheyakale:
Should we have 2 different URL patterns in yaml for otp and api keys rather than the same pattern? Although technically it doesn't make a difference, accidental deletion of api key would be less desirable as against deletion of otp keys. Other than that, the code seems to be readable and good to merge.

sanand0:
You're right -- in production, we would likely have different URLs. Or NOT expose any URL for revocation.

The URL `/auth/revoke` is only for test cases. I combined OTP and API Key revocation into a single URL just to reduce test code.

Will merge. Thanks!

PR #588, opened 2022-08-08
DEP: Remove deprecated rmarkdown FileHandler template

sanand0:
RMarkdown is rarely used with Gramex. It is very hard to support.
The underlying packages have several installation issues and
compatibility issues with NumPy. So we're phasing out support.

We deprecated it in v1.81 and are removing it in v1.83.


jaidevd (approved):

radheyakale:
Looks good to merge.

PR #600, opened 2022-09-14
ENH: MLPredictor runs ML predictions on FormHandler datasets

jaidevd:
MLPredictor takes a FormHandler-like dataset spec and makes prediction on each row, given a sklearn model. FormHandler filters can be applied to datasets as usual.

### DIFF
--- a/gramex/handlers/mlhandler.py
+++ b/gramex/handlers/mlhandler.py
@@ -13,7 +13,6 @@ from gramex.http import NOT_FOUND, BAD_REQUEST
 from gramex.install import safe_rmtree
 from gramex import cache
 
-from orderedattrdict import AttrDict
 import numpy as np
 import pandas as pd
 import joblib
@@ -362,14 +361,6 @@ class MLPredictor(FormHandler):
         kwargs.update(data)
         super(MLPredictor, cls).setup(**kwargs)
 
-    @coroutine
-    def get(self, *path_args, **path_kwargs):
-        for key, dataset in self.datasets.items():
-            break  # We assume that unlike in FormHandler, only one dataset is present
-        opt = self._options(dataset, self.args, path_args, path_kwargs, key)
-        meta = {key: AttrDict()}
-        data = yield gramex.service.threadpool.submit(
-            self.data_filter_method, args=opt.args, meta=meta[key], **opt.filter_kwargs)
-        dataset.pop('schema', None)
-        result = yield gramex.service.threadpool.submit(self.model.predict, data)
-        self.write(json.dumps(result, cls=CustomJSONEncoder))
+    def modify_all(self, data=None, key=None, handler=None):
+        df = pd.DataFrame.from_dict({'prediction': self.model.predict(data['data'])})
+        return {'data': df}
--- a/gramex/handlers/mlhandler.py
+++ b/gramex/handlers/mlhandler.py
@@ -356,11 +356,11 @@ class MLHandler(FormHandler):
 class MLPredictor(FormHandler):
 
     @classmethod
-    def setup(cls, model, data, **kwargs):
-        cls.model = cache.open(model, joblib.load)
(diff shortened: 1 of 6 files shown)


sanand0:
You could instead subclass FormHandler like this:

```python
class MLPredictor(FormHandler):
    def modify_all(self, data=None, key=None, handler=None):
        self.model.predict()
        return any_dict_that_will_be_rendered
```

jaidevd:
@sanand0 Done - please take a look.

tests/gramex.yaml
```diff
@@ -1279,6 +1279,14 @@ url:
         class: NER
       xsrf_cookies: false
 
+  mlpredictor:
+    pattern: /predictor
+    handler: MLPredictor
+    kwargs:
+      model: $YAMLPATH/predictor.pkl
+      data:
```
sanand0, line 1287: Like FormHandler, allow `url:` to be directly under kwargs since we don't (yet) support multiple datasets. (And even if we do, we can handle it just like FormHandler)

Similarly, use $YAMLPATH/predictor/model.pkl

gramex/handlers/mlhandler.py
```diff
@@ -351,3 +351,16 @@ def delete(self, *path_args, **path_kwargs):
                 getattr(self, f'_delete_{item}')()
             except AttributeError:
                 raise HTTPError(BAD_REQUEST, f'Cannot delete {item}.')
+
+
+class MLPredictor(FormHandler):
+
+    @classmethod
+    def setup(cls, model, data, **kwargs):
```
sanand0, line 359: `data` does not need to be a parameter. We will allow `url:` etc at the top level.

Let's replace `model` with `config_dir` and ask users to place their `.pkl` under `config_dir` as `model.pkl`. Makes it consistent with MLHandler. (Might mean more work for the user, but let's favor consistency for now, and simplify it for users with another config later if required.)

gramex/handlers/mlhandler.py
```diff
@@ -351,3 +351,16 @@ def delete(self, *path_args, **path_kwargs):
...
+        cls.model = cache.open(model, joblib.load)
+        kwargs.update(data)
```
sanand0, line 361: We can drop this since `data:` is no longer required.

tests/utils.py
```diff
@@ -497,6 +499,16 @@ def transform_circles(df, *argss, **kwargs):
     return df
 
 
+def make_classifier():
```
sanand0, line 502: Let's move this to `setupClass` in TestPredictor

gramex/handlers/mlhandler.py
```diff
@@ -351,3 +351,16 @@ def delete(self, *path_args, **path_kwargs):
                 getattr(self, f'_delete_{item}')()
             except AttributeError:
                 raise HTTPError(BAD_REQUEST, f'Cannot delete {item}.')
+
+
+class MLPredictor(FormHandler):
+
+    @classmethod
+    def setup(cls, model, data, **kwargs):
+        cls.model = cache.open(model, joblib.load)
```
sanand0, line 360: Let's load the model on demand, AND via `gramex.cache.open` -- so if the file changes, we reload the model

tests/gramex.yaml
```diff
@@ -1279,6 +1279,13 @@ url:
         class: NER
       xsrf_cookies: false
 
+  mlpredictor:
+    pattern: /predictor
+    handler: MLPredictor
+    kwargs:
+      config_dir: $YAMLPATH
+      url: $YAMLPATH/../testlib/iris.csv
+
```
sanand0, line 1288: Apologies -- let's revert to:

```yaml
    data:
        url: ...
```

This is consistent with MLHandler.

Also, let's support an optional `columns:` parameter that specifies the columns in case we use a non-MLHandler based .pkl file

PR #604, opened 2022-09-26
Dissolve a for loop that concatenate strings to str.join

boneyag:
Line 37, 38 of the file `tests/functionutils.py` uses a loop to concatenate strings. 
I think the more pythonic way is to use `str.join` instead. 
On the performance level, `str.join` is much faster if the list is very long.


sanand0:
Good catch. Thanks, @boneyag

PR #631, opened 2022-11-02
BLD: Remove conda package. Use pip + npm, or Docker

sanand0:
@jaidevd @bhatsandeep I propose we keep our installation options to just the following:

1. `pip install gramex && gramex setup --all`
2. Docker

Since we've deprecated R, and are decoupling node versions from conda, there's no reason to maintain a conda package. We can run `pip install gramex && gramex setup --all` inside conda.

This PR removes the conda build infrastructure. Thoughts?


jaidevd:
This means users will have to install node on their own. Otherwise this is fine.

PR #653, opened 2022-11-23
BLD: Compute Gramex feature size & complexity

sanand0:
This creates a `reports/gramexsize.csv` that shows cyclomatic complexity of Gramex features.

`gramexsize.yaml` has a `code2yaml`. It maps

- from Python wildcard paths (e.g. modules like `gramex.cache.*` or methods like `gramex.handlers.basehandler.BaseMixin.setup`)
- to YAML wildcard paths (e.g. `url.*.handler=FormHandler|FilterHandler|MLHandler|DriverHandler`)

It calculates the complexity of the Python paths, and saves it along with the YAML paths in `reports/gramexsize.csv`.

In a project, we can match the YAML paths used, and calculate the corresponding Python complexity. This gives us the complexity of the Gramex features used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ tests/watcher.txt
 *.csv
 *.xls*
 
+!reports/*.csv
 !tests/**/*.csv
 !tests/*.xlsx
 !testlib/**/*.csv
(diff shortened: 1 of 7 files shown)


sanand0:
@jaidevd @bhatsandeep -- would like your point of view on the `gramexsize.yaml` mappings, please. The TODOs as well as what I have currently.

PR #671, opened 2023-01-03
ENH: gramex.data.* functions accept argstype to typecast args

sanand0:
- ENH: argstype: converts URL args to Python types in SQL params
- ENH: argstype: supports expanding=True to bind iterable SQL params
- API: gramex.data.{update,delete,insert} use positional args before meta
- REF: data.{filter,update,delete,insert} have consistent signatures
- REF: Use data._convertor() for all internal type conversions
- REF: gramex.cache uses a single JSON dump with CustomJSONEncoder
- FIX: support filtering string columns that has NaN/NULL
- REF: improve type checking

### DIFF
--- a/gramex/cache.py
+++ b/gramex/cache.py
@@ -435,6 +435,13 @@ def stat(path: str) -> Union[Tuple[float, int], Tuple[None, None]]:
     return (None, None)
 
 
+def _json_dump(obj: Any) -> str:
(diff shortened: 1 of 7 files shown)


bhatsandeep (approved):
Changes look good to me. Though too many for a single commit. :)

PR #680, opened 2023-01-11
Ss setup scripts

shraddheya:
Creating Installer script that supports ubuntu, centos and Amazon Linux 2
Added test cases for all three OS.


sanand0:
Closed as outdated

PR #683, opened 2023-01-15
BLD: gramex available in PATH of Docker image

sanand0:


jaidevd (commented):
@sanand0 is it possible to activate the conda env instead of only adding Gramex to $PATH? We might need access to more than just the Gramex exe.

I'm currently in the process of getting the IDE to use this official image instead of the Ubuntu based one we had earlier. Activating the conda env will make it a lot easier for me.

sanand0:
Activating the conda env sets up a [set of things (e.g. prompt, new bash functions)](https://stackoverflow.com/a/48591320/100904) that are specific to the shell, and I believe we can't inherit those without explicitly running `conda activate` IN the shell.

That's why I just added the PATH. But if there's anything additional I can do that will persist for child images using this one, I'd definitely want to. Do let me know.

Issue #687, opened 2023-01-17
Allow user, alert and request logging to be distributed


sanand0:
Currently, user login is stored in `$GRAMEXDATA/logs/user.csv` as a TimedRotatingFileHandler. Alerts are similarly stored in alert.csv and requests in request.csv.

This causes issues with multi-instance deployments when the rotations may conflict with each other.

Migrate these to databases under `storelocations`.

sanand0:
To migrate user login:

- [ ] Add a `storelocation` for user logs
- [ ] Create a `log_user_event` function that stores in this user log as well as the existing storage location
- [ ] Deprecate the existing storage location for 1 Jan 2024 and schedule removal

sanand0:
Closed via #707

PR #694, opened 2023-02-13
ENH: storelocations.userlog stores user actions. #687

sanand0:


sanand0:
TODO: Allow developer to pick max records or max retention date.

PR #728, opened 2023-06-01
Roles and Permissions for Gramex SAAS

shraddheya:
# ADD Roles and permissions model to be used in SAAS.

Adds a new database in `storelocations` by the name of `roles.db` with tables

- roles: `roles` and `users` mapping
- permissions: `permissions` and `roles` mapping 
- user_permissions: mapping for extra permissions that can be assigned to some `user`

This module adds two new fields in `attrdict` of `handler.current_user` and `session['user']` name:

- roles
- permissions


sanand0:
Looks mostly good, @shraddheya. Let's discuss and test it out over a call? Thanks!

PR #729, opened 2023-06-01
Gramex tasks

shraddheya:
Schedule detached long running background tasks in gramex


sanand0:
A few suggestions @shraddheya 

- [ ] Could you please plan a Python API that aligns with the command line API
- [ ] Could you please add documentation as docstrings
- [ ] Could you please set it up so that we can independently push this on PyPi as gramextask or the like

PR #733, opened 2023-06-05
TST: MessageHandler test cases with pytest

sanand0:
This PR does 3 things:

1. When `bash task test` runs, it runs `gramex` on port 9999 and then runs pytest, allowing it to test handlers
2. We use this to run tests for MessageHandler in pytest
3. `.gitlab-ci.yml` allows running these tests on code.gramener.com


sanand0:
@bhatsandeep @jaidevd -- how could I make this PR clearer, please?

jaidevd:
> When bash task test runs, it runs gramex on port 9999 and then runs pytest, allowing it to test handlers

Does this mean that this is the new way of running tests, as against the earlier way of simply running `$ nosetests`?

gramex/handlers/__init__.py
```diff
@@ -93,6 +88,17 @@
...
+
+# MLHandler requires optional dependencies scikit-learn and statsmodels.
+# If they're not installed, ignore the import error and skip MLHandler.
+try:
+    from .mlhandler import MLHandler, MLPredictor
+
+    ML = MLHandler
+    __all__ += ['ML', 'MLHandler', 'MLPredictor']
+except ImportError:
+    pass
```
jaidevd, line 100: Please log a warning.

jaidevd:
Also, what's the deal with the new folder `pytest`? Why isn't all of this within the `tests/` directory as usual?

sanand0:
@jaidevd 

>> When bash task test runs, it runs gramex on port 9999 and then runs pytest, allowing it to test handlers
> 
> Does this mean that this is the new way of running tests, as against the earlier way of simply running $ nosetests?

Yes. nosetests used to run `gramex`. Now, we run `gramex` and THEN run `pytest`. Advantage: clean shutdown of the tests. (nosetests sometimes doesn't cleanly shutdown because of some threading nightmare.)

> Also, what's the deal with the new folder pytest? Why isn't all of this within the tests/ directory as usual? Seems like we're deviating significantly from the older way of testing Gramex.

I first created a `tests` folder with nosetests. It started `gramex` and ran tests. But startup was slow.

I realized we don't need to run gramex for APIs like gramex.cache.open, etc. So I created an independent folder of fast tests called `testlib`.

Now, as we migrate to `pytest`, I can't use the same `tests` folder because pytest also ends up running the nosetests test cases (and messes it up.)

So we use `nosetests` to run the `testlib` and `tests` folder but will eventually migrate to `pytest`. Once done, I'll rename the `pytest` folder to `tests`. Maybe in 2025 :-(

jaidevd:
> Now, we run gramex and THEN run pytest.

This is great...
It would be good to have a sample `.travis-ci.yml` script here which shows how to run the _whole_ testsuite, including pytest, tests, testlib, etc.

PR #758, opened 2023-07-03
ENH: Add sentry service

sanand0:
Gramex integrates with [Sentry](https://sentry.io/) out-of-box.

First, [log into Sentry, create a project](https://docs.sentry.io/product/sentry-basics/integrate-frontend/create-new-project/) and copy the DSN key. 

Then, add this to `gramex.yaml`:

```yaml
sentry:
  dsn: "https://xxx@xxx.ingest.sentry.io/xxx"  # paste your DSN key
```

When Gramex reports any errors on the console, these are automatically sent to your [Sentry](https://sentry.io) instance. This works with [self-hosted sentry](https://develop.sentry.dev/self-hosted/) too.

Gramex adds the following information to each error:

1. The `gramex` tag set to the Gramex version (e.g. `1.92.0`)
2. The `cwd` tag is set to the current working directory (e.g. `C:\temp\`)
3. The `port` tag is set to the port Gramex is running on (e.g. `9988`)
4. The `user` is set based on the current user object (e.g. `{"id": "user@example.org", "role": "..."}`

You can pass [options to configure Sentry](https://docs.sentry.io/platforms/python/configuration/), like:

```yaml
sentry:
  dsn: "https://xxx@xxx.ingest.sentry.io/xxx"  # paste your DSN key
  sample_rate: 0.1
  server_name: my-server.example.com
  environment: production
```

Currently, only scalar options (text, numbers, booleans) are supported, not functions.

### DIFF
--- a/gramex/handlers/basehandler.py
+++ b/gramex/handlers/basehandler.py
@@ -5,6 +5,7 @@ import json
 import time
 import logging
 import mimetypes
+import sentry_sdk
 import traceback
 import tornado.gen
 import gramex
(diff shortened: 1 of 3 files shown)


sanand0:
Closed as outdated

